PR #45, opened 2024-10-05
Support docker (:

crabest:
This will help people to dockerize and self host postgres quickly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,4 +40,3 @@ next-env.d.ts
 # docker
 .docker
 docker-compose.override.yml
-docker-compose.yml
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,17 @@
+FROM node:20-alpine
+
+WORKDIR /app
+
+RUN npm install -g pnpm
+
+COPY package.json pnpm-lock.yaml* ./
+
+RUN pnpm install
+
+COPY . .
+
+RUN pnpm build
+
+EXPOSE 3000
+
+CMD ["pnpm", "start"]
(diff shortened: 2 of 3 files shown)


wiksien:
I don't think this PR fits well with the current dev setup workflow. This code requires the user to already have a created .env file with all of the environment variables in it. Of course, fresh devs don't have .env files yet, so there is the setup.ts file to help them out. To run setup.ts you need Node, so there is no need to run a separate Docker node instance.

The setup.ts itself has an option to run Postgres through Docker with docker-ompose.yml, so if this PR goes through, now we have two versions of the docker-compose.yml file. One from the repo itself, one from the code in setup.ts.

Also, I think the command `pnpm start` in the Dockerfile should be replaced with `pnpm dev` since `pnpm start` creates a production app, not dev.

Big thank you for the effort you put into this PR. Anyone feel free to disagree with me and add a comment. :)

wiksien:
In my opinion the way this PR could be integrated into the current workflow perhaps would be to create a dockerfile that runs a node instance that then runs the setup file that sets up all the other stuff and creates the .env file and then runs the other 2 db commands needed for drizzle.

Oh, also the README.md needs to be updated to inform the dev about this option of setup.

slavingia:
Why does one need docker at all? I'd love to not add this level of complexity.

wiksien:
I agree with Slavingia. Docker currently is just a side option to host postgres local (main way is Vercel). Docker adds complexity for little benefit in my opinion.

crabest:
@wiksien thank you for your detailed feedback, I will try to work on the part of setup.ts to see if any improvements to add.
Also Docker to host this application on their own linux server that's why I did use `pnpm start` instead of `pnpm dev`.
I will focus now more about features then.

PR #48, opened 2024-10-06
feat: connect linear issues

juniusfree:
## Problem

When I write unit tests without the right context, they can end up inaccurate or incomplete. Missing key details from issues or tickets  means the tests might not cover important parts of the feature. This leads to gaps in coverage, increasing the risk of bugs making it into production.

## Feature

To solve this, I added a feature that connects with Linear. Users can link their Linear accounts, search for related issues directly in the app, and add them as context when generating unit tests. This extra context makes the tests more accurate and relevant, ensuring they reflect the feature requirements.

## Scope Limitations

This version is a proof of concept, focusing on Linear integration to validate its effectiveness and see if it helps other developers. If it proves useful, expanding to include other tools like Jira and adding more integrations will be the next steps.

## Demo

https://github.com/user-attachments/assets/94d9c7f4-9dcb-4abf-bb7e-9c105caf14a2

### DIFF
--- a/app/(dashboard)/dashboard/issue-modal.tsx
+++ b/app/(dashboard)/dashboard/issue-modal.tsx
@@ -1,0 +1,152 @@
+"use client";
+
+import { Button } from "@/components/ui/button";
+import {
+  Dialog,
+  DialogContent,
+  DialogDescription,
+  DialogHeader,
+  DialogTitle,
+} from "@/components/ui/dialog";
+import { searchIssues } from "@/lib/linear";
+import { useClerk } from "@clerk/nextjs";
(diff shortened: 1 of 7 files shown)


slavingia:
I think the code should be self-explanatory to writing the tests.

slavingia:
I'm open to a github issue integration perhaps, or feeding in the PR description, but not Linear / Notion as I think those tools shouldn't contain "specs" outside of GitHub and limits the usage. 

We can discuss on the call tmmrw tho!

slavingia:
Link in discussions! Will also be recorded

PR #64, opened 2024-10-19
Improve streaming test file content generated by AI

m2rads:
Added a generator function `streamTestFiles` to progressively render AI-generated test files. This will improve streaming test files that was originally implemented with useObject. 

Key changes:
- Uses a generator to yield file content in chunks
- Configurable chunk size (currently 70 characters)
- Streams multiple files sequentially
- Provides file index and content chunk for each yield

This enhancement allows for smoother rendering of large test files, reducing perceived load times and improving overall user experience.

This is basically stimulating streaming. In practice we should implement streaming in server side.

### DIFF
--- a/app/(dashboard)/dashboard/pull-request.tsx
+++ b/app/(dashboard)/dashboard/pull-request.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import { useCallback, useState } from "react";
+import { useState } from "react";
 import { Button } from "@/components/ui/button";
 import {
   GitPullRequestDraft,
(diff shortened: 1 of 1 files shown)


app/(dashboard)/dashboard/pull-request.tsx
```diff
@@ -70,6 +70,9 @@ export function PullRequestItem({ pullRequest: initialPullRequest }: PullRequest
   const { toast } = useToast();
   const [commitMessage, setCommitMessage] = useState("Update test files");
 
+  const [streamedTestFiles, setStreamedTestFiles] = useState<TestFile[]>([]);
```
slavingia, line 73: This seems a bit boilerplate-y and I think we'd get this for free with streamObjects?

A video comparison would also help

m2rads: I totally agree with you and this is not my preferred way of doing it. I think streaming should happen in the generate-test route and handle it in pull-request. However, I temporarily made these changes because the current implementation is glitchy and in some cases, it fails to handle exceptions properly and causes the app to crash (unfortunately was not able to show this in the video but I faced this issue a couple times). 

This is boilerplate-y but it's reliable, however, if you don't feel comfortable with this, it would be better to revert streaming test files for now in my opinion. 

Quick side by side comparison (left one being this branch) 

https://github.com/user-attachments/assets/70260b7a-f72b-45b0-86d4-da7552e58a85

m2rads:
I will close this for now. I observed some issues after resolving merge conflicts.

PR #72, opened 2024-10-26
Add quick switch among models from openai, anthropic, google (like in cursor)

poudelprakash:
screenshot:
image

### DIFF
--- a/app/(dashboard)/dashboard/pull-request.tsx
+++ b/app/(dashboard)/dashboard/pull-request.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import { useState, useCallback } from "react";
+import { useState, useCallback, useEffect } from "react";
 import { experimental_useObject as useObject } from "ai/react";
 import Link from "next/link";
 import dynamic from "next/dynamic";
@@ -22,6 +22,7 @@ import {
 import { Button } from "@/components/ui/button";
 import { Checkbox } from "@/components/ui/checkbox";
 import { Input } from "@/components/ui/input";
+import { Select } from "@/components/ui/select";
 import { useToast } from "@/hooks/use-toast";
 import {
   commitChangesToPullRequest,
@@ -57,6 +58,25 @@ export function PullRequestItem({
   const [error, setError] = useState<string | null>(null);
   const [commitMessage, setCommitMessage] = useState("Update test files");
   const { toast } = useToast();
+
+  const { data: availableModels, isLoading: modelsLoading } = useSWR(
+    'api/ai-models',
+    async () => {
+      const res = await fetch('/api/ai-models');
+      if (!res.ok) {
+        throw new Error('Failed to fetch AI models');
+      }
+      return res.json();
+    }
+  );
+
+  const [ai_model, setAiModel] = useState(availableModels ? availableModels[0] : null);
+
+  useEffect(() => {
+    if (availableModels && availableModels.length > 0) {
+      setAiModel(availableModels[0]);
+    }
+  }, [availableModels]);
 
   const { data: pullRequest, mutate } = useSWR(
     `pullRequest-${initialPullRequest.id}`,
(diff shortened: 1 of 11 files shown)


poudelprakash:
This PR is ready for review

app/(dashboard)/dashboard/pull-request.tsx
```diff
@@ -22,6 +22,7 @@ import {
 import { Button } from "@/components/ui/button";
 import { Checkbox } from "@/components/ui/checkbox";
 import { Input } from "@/components/ui/input";
+import { Select } from "@/components/ui/select";
```
slavingia, line 25: IMO this should be in a new Settings page instead of per PR, or next to the filters.

slavingia (changes requested):

app/(dashboard)/dashboard/pull-request.tsx
poudelprakash: something like cursor?
ability to provide api keys for each provider,
enable disable each model
ability to add model as string

and set one mode as default/or choose only one model among availabel.

image

more guidance on how the feature would like like from your perspective will help.

lib/ai-models.ts
```diff
@@ -0,0 +1,49 @@
...
+  const getAiModels = () => {
+    const models = [];
+
+    if (process.env.ANTHROPIC_API_KEY) {
+      models.push(...Object.keys(modelProviderMap).filter((model) => model.includes("claude")));
+    }
+    if (process.env.GOOGLE_GENERATIVE_AI_API_KEY) {
+      models.push(...Object.keys(modelProviderMap).filter((model) => model.includes("gemini") || model.includes("gemma")));
+    }
+    if (process.env.OPENAI_API_KEY) {
```
poudelprakash, line 43: @slavingia do you approve of this approach to reference api key from env file to see which models to provide as choice

or would saving api keys in db be the way?

app/(dashboard)/dashboard/pull-request.tsx
poudelprakash: also my perspective here was as a developer, it was easy way for me to compare test generation from different models when this option is in per item,
I was thinking of capability to set default model for account though

I can understand, as a user this might not be the thing one want to select or think about each time they generate the tests..

poudelprakash:
@slavingia is there any reference page for settings page?
there might be tabs inside the setting page and form should probably look certain way..

what would be best way to plan to move this to settings page?

app/(dashboard)/dashboard/pull-request.tsx
slavingia: Think that's not keeping in shortest config personally. Seems like part of the value is having shortest make those changes to always use SOTA

poudelprakash: @juniusfree in my opinion, shortest.yml should not have any repo specifications, current app lets us connect our github account and work with any repo, 
having repo config in yml seems like pulling away that functionality.
Let me know if I understood it wrong.

Shortest currently does not have repo listing page or per repo config ui, something like codeclimate's ui for per repo config would probably be better place to manage per repo config.
image

I think for current use case, being ability to select preferred model for account would be enough.
If there were some initial settings page design, it would be easier for me to continue(I assume core-team would have highly opinionated designs for such new pages)

poudelprakash:
@slavingia can you please provide some action item for this pr, so I can take this to finish line?

slavingia:
Will just close. Shortest should just determine the best model at the time, not something the user should have to think about

poudelprakash:
@slavingia Can I please have proper feedback on pr at the very least?

‘shortest should determine the best model at the time’ is kind of hard to understand..
wouldn’t such mechanism require way to switch ai models?

even though the select dropdown might not be in right place and it can be changed to other place,
the pr already has code to add the ability to switch AI models.

I am not paid to do this work,
at least a feedback I can understand would be very good return for my hours of effort.

slavingia:
Sure, I mean that there's only ever going to be one model at a time that performs the best, and it should be hardcoded. We don't need to support multiple and switching at all. Right now it's seen to be Claude 3.5 Sonnet New

PR #77, opened 2024-10-28
fix: improve GitHub Actions detection messaging

bjornpagen:
### Problem
When someone starts using shortest for the first time, they may not realize shortest actually reads their build logs to try to autocorrect tests. Currently, if no test runners are detected by shortest, it fails silently without notifying the user.

### Solution:
Clarify to users that GitHub Actions is needed for test log analysis by adding PR number and repo details to the error message when Actions is not enabled. This helps users understand why Actions access is required for the test-fixing functionality to work properly.

Screenshot 2024-10-27 at 8 26 38 PM

### DIFF
--- a/app/(dashboard)/dashboard/pull-request.tsx
+++ b/app/(dashboard)/dashboard/pull-request.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import { useState, useCallback } from "react";
+import { useState, useCallback, useEffect } from "react";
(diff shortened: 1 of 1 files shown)


m2rads:
So this issue happens even if you have workflow and actions:read scope added in Clerk? 

image

slavingia:
Good point, can't know if tests are failing without GitHub actions

PR #79, opened 2024-10-29
security: prevent db queries from being callable on the client.

bjornpagen:
This PR changes `queries.ts` from "use-server" to "server-only", ensuring that DB query functions are reserved for server-side use.

Exported functions in a `"use-server"` file are exposed to the frontend. This meant users could call functions like `getUserByClerkId()` or `updateUserSubscription()` with an arbitrary `clerkId`.

https://nextjs.org/blog/security-nextjs-server-components-actions#write

Impact is low, since `clerkId` is a random value.

### DIFF
--- a/lib/db/queries.ts
+++ b/lib/db/queries.ts
@@ -1,4 +1,4 @@
-"use server";
+import "server-only"
 
 import { eq } from "drizzle-orm";
 import { db } from "./drizzle";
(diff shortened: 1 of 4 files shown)


slavingia:
You're referencing a year-old blog post. Don't think this is encouraged. "use server" should work fine.

slavingia:
If truly a vulnerability, could you replicate?

PR #85, opened 2024-11-21
Feature: Implement Assertion 

m2rads:
Add assertion support and DB validation to test framework

Changes:
- Added Jest expect integration for assertions
- Added support for async DB validations in test steps
- Improved error handling and reporting for assertions
- Added test script to verify assertion implementation
- Updated test builder to handle both sync/async test execution

Example usage:

```typescript
.when('Logged in', async () => {
const user = await db.query.users.findFirst();
expect(user).toBeDefined();
})
```

Demo: 
image

PR #88, opened 2024-11-23
Feature: AI Executes Callback functions

m2rads:
This feature covers AI executing callback functions in the test suite.

Why is this feature important? 

This feature allows the callback functions to be executed at the right time. For example there might be some test logics that should only be executed after a certain event has been executed by AI. This is now possible because AI takes control of executing callback functions. 

This is also useful when users want to validate a 3rd party provided in their application but shortest does not provide support for that. This way developers can write their own third party validation and mix is with AI driven e2e tests. 

For example test case please look at login.test.ts file. 
Example output: 
Screenshot 2024-11-23 at 1 00 37 PM


m2rads:
Just a note that this PR depends on PR #85. Please merge that one first.

slavingia:
Merged! Some new merge conflicts have come up.

m2rads:
Resolved conflicts.

PR #92, opened 2024-11-28
fix pnpm version mismatch

m2rads:
Publish workflow failed because of pnpm version mismatch


slavingia:
^ if you're in the Vercel team, that error should stop happening no?

PR #93, opened 2024-11-28
README Update

m2rads:
Just a quick update to shortest package to enforce publish job to run.


m2rads:
Yes so currently I setup the publish workflow in a way that it would only run when a new change has been applied to packages/shortest/**. There was an error in publish that I fixed. Now I just made a dummy change in README to make the job run again. Just troubleshooting the workflow to make sure it publishes the package properly.

m2rads:
Sorry my bad. I totally forgot that we can manually run the workflows. Closed this one.

PR #95, opened 2024-11-29
Bug: Global types are not registered in typescript properly

m2rads:
fix: global type declarations

Fixed TypeScript global declarations for `define`, `expect`, `beforeAll`, and `afterAll` functions. Updated type exports to avoid duplication and ensure proper type resolution in consuming projects.

Changes:
- Added proper global declarations in index.d.ts
- Removed duplicate type exports from index.ts
- Fixed build script for type declarations

Bump version to 0.0.2 and update changelogs


m2rads:
Hi @slavingia I will just merge this one because it's a bit of an urgent fix. Hope it's ok with you.

PR #97, opened 2024-11-29
Update README.md

slavingia:
To help make it clear how to use Shortest.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,38 @@
+# Quick start
+
+1. Install the Shortest package
+   
+```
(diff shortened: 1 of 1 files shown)


slavingia:
Could probably still be clearer, e.g what to name the file and where to place it and how does shortest determine the route (it always starts from / or does it use the folder it's in to determine it "magically"?)

m2rads:
For now it uses the testDir property in shortest.config.ts to find the test file. Maybe we could add a new section for configurations.

image

PR #98, opened 2024-11-30
Fix Lifecycle hooks and improve browser navigation context

m2rads:
Version 0.0.3

### Fixed
- Fixed execution order of lifecycle hooks
- Fixed CLI help command requiring GitHub TOTP secret
- Improved browser navigation performance using 'load' instead of 'networkidle'
- Fixed GitHub tool initialization to be lazy-loaded
- Improved error handling in browser navigation

### Changed
- Reduced navigation timeouts for better performance
- Made GitHub TOTP validation more flexible
- Improved browser cleanup on process termination


packages/shortest/src/ai/client.ts
```diff
@@ -99,9 +99,7 @@ export class AIClient {
             .filter(block => block.type === 'tool_use')
             .map(block => {
               const toolBlock = block as Anthropic.Beta.Messages.BetaToolUseBlock;
-              if (toolOutputCallback) {
-                toolOutputCallback(toolBlock.name as string, toolBlock.input);
-              }
+              // console.log('🛠️ Executing tool:', toolBlock.name, 'with input:', toolBlock.input);
```
slavingia, line 102: Should remove or keep, not a fan of comments like this

m2rads:
@slavingia comments removed

packages/shortest/src/index.ts
```diff
@@ -119,18 +129,63 @@ export class TestRegistry {
...
   static clear() {
     global.__shortest__.registry.suites.clear();
     global.__shortest__.registry.currentSuite = null;
-    global.__shortest__.registry.beforeAllFns = [];
-    global.__shortest__.registry.afterAllFns = [];
+    this.suiteHooks.clear();
+  }
+
+  static getTestBuilder(test: TestCase): UITestBuilder | null {
+    // console.log('🔍 Looking for builder:', {
```
slavingia, line 170: Can we remove these comments too?

packages/shortest/src/index.ts
```diff
@@ -119,18 +129,63 @@ export class TestRegistry {
...
+    //   suiteName: test.suiteName,
+    //   testName: test.testName
+    // });
+    
+    const suite = this.suites.get(test.suiteName);
+    // console.log('🔍 Found suite:', suite ? 'yes' : 'no');
```
slavingia, line 176: Here

packages/shortest/src/index.ts
```diff
@@ -119,18 +129,63 @@ export class TestRegistry {
...
+    
+    if (!suite) return null;
+    
+    const builder = suite.find((builder: UITestBuilderInterface) => 
+      builder.testName === test.testName
+    );
+    
+    // console.log('🔍 Found builder in suite:', builder ? 'yes' : 'no');
```
slavingia, line 184: Here

slavingia (approved):

packages/shortest/src/index.ts
m2rads: Sorry missed these. Done!

app/__tests__/assert.test.ts
```diff
@@ -8,7 +8,19 @@ interface User {
 
 define('Test Assertions', async () => {
   // Test 1: Basic Assertions (Will Pass)
+  afterAll(async () => {
+    console.log('Clearing DB after all tests');
+  });
+  beforeAll(async () => {
+    console.log('Clearing DB before all tests');
+  });
    new UITestBuilder<User>('/')
```
slavingia, line 17: Would be great to remove this, it's quite confusing to a user what new UITestBuilder is.

slavingia: The README shows what every test should look like:

```
import { define } from 'shortest'

define('Login Flow', () => {
  test('User can login')
    .given('I am on the login page')
    .when('I enter my credentials')
    .expect('I should be logged in')
})
```

The fewer lines of code and chars required the better, as it's in the name!

slavingia: IMO we could shorten further:

import { test } from 'shortest'

test('the user can login');

AI should be able to infer the rest.

m2rads: I absolutely agree. We should make it more intuitive and we can start by removing extra stuff. maybe we should just have test('', ()=>{}) and an optional callback function in case some operations need to happen in between. I will make this improvement.

PR #100, opened 2024-12-04
Feat: Making Playwright's page object available in Test Files

m2rads:
Mocking logging in is now possible: 

```typescript
await clerk.signIn({
        page,
        signInParams: { 
          strategy: "email_code", 
          identifier: "iffy+clerk_test@example.com" 
        },
 });
```

PR #101, opened 2024-12-04
Improve home.test.ts

slavingia:


app/__tests__/home.test.ts
```diff
@@ -1,26 +1,7 @@
-import { UITestBuilder } from '@antiwork/shortest';
+import { test } from '@antiwork/shortest';
```
slavingia, line 1: We have this in the README, not sure why this doesn't match.

app/__tests__/home.test.ts
```diff
@@ -1,26 +1,7 @@
...
 
 interface loginButton {
   url: string;
 }
 
-define('Validate Dasboard is accessible by users', async () => {
-  afterAll(async () => {
```
slavingia, line 8: We shouldn't have random console logs that don't test anything.

app/__tests__/home.test.ts
```diff
@@ -1,26 +1,7 @@
...
-    console.log('Clearing DB after all tests');
-  });
-  beforeAll(async () => {
-    console.log('Clearing DB before all tests');
-  });
-  new UITestBuilder<loginButton>('/')
```
slavingia, line 14: Removed UITestBuilder, too complex. Also shouldn't need to have an Interface created. Humans don't think this way.

app/__tests__/home.test.ts
```diff
@@ -1,26 +1,7 @@
...
-    .test('Validate that users can access the dashboard')
-    
-    .given('baseUrl', { url: 'http://localhost:3000' })
```
slavingia, line 17: Should default to /, and otherwise be in the proper folder with a page.tsx or route.ts and then the route can be assumed due to filesystem-based routing of Nextjs. we could add a url option for tests that don't follow that format

(in the case of Shortest, we should move the tests to those places so it can serve as a really good demo.)

app/__tests__/home.test.ts
```diff
@@ -1,26 +1,7 @@
...
-    .when('Clicking on view dashboard button')
-    .expect('Should be able redirect to /dashboard and see the dashboard')
```
slavingia, line 19: The AI can assume the test has passed by being able to see the dashboard via computer use, so this isn't necessary IMO.

slavingia (commented):
Updated to the "ideal" so that we have something clear to work towards and avoid miscommunication about the goal of Shortest going forward.

app/__tests__/home.test.ts
m2rads: Oh! I updated the the README with the correct syntax.

app/__tests__/home.test.ts
m2rads: Yes we're removing them.

app/__tests__/home.test.ts
m2rads: Removed the logs in the same PR #100

m2rads:
Finally got the new design to work. As we discussed here are some examples: 

shortest:
```typescript
import { test } from '@antiwork/shortest';
import { db } from "@/lib/db/drizzle";
import { sql } from "drizzle-orm";
import dotenv from "dotenv";

dotenv.config();

const githubCredentials = {
  email: process.env.GITHUB_EMAIL,
  username: process.env.GITHUB_USERNAME,
  password: process.env.GITHUB_PASSWORD,
}

test('Login to the app using Github login', githubCredentials , async ({ page }) => {
    console.log('Validating DB for the given user: ', githubCredentials.username);
    
    try {
      console.log('Starting DB validation...');
      const [customer] = await db.execute<{ id: string, name: string, email: string }>(sql`
        SELECT * FROM customers WHERE email = 'argo.mohrad@gmail.com'
      `);
  
      if (!customer) {
        throw new Error('Customer argo.mohrad@gmail.com not found in database');
      }
  
      console.log('Found customer in DB:', customer);
      expect(customer.email).toBe(githubCredentials.username);
      expect(customer.name).toBe(githubCredentials.email);
  
    } catch (error) {
      console.error('DB Validation Error:', error);
      throw error;
    }
})

test('Validate write new test button generates test cases and make sure you can commit new tests to github')
```
iffy: 

```typescript
import { test } from '@antiwork/shortest';
import { clerk, clerkSetup } from "@clerk/testing/playwright";

test.beforeAll('Setup Clerk', async () => {
    console.log('Setting up Clerk');
  await clerkSetup({
    frontendApiUrl: process.env.PLAYWRIGHT_TEST_BASE_URL ?? "http://localhost:3000",
  });
});

test.beforeEach(async ({ page }) => {   
    console.log('Setting up Clerk before each test'); 
    try {
        await page.goto("http://localhost:3000/dashboard", { 
          waitUntil: 'networkidle',
          timeout: 3000000
        });
        
        await clerk.signIn({
          page,
          signInParams: { 
            strategy: "email_code", 
            identifier: "iffy+clerk_test@example.com" 
          },
        });
    } catch (error) {
        console.error('Authentication error:', error);
    }
});

test('Validate that title "Moderation" is visible')
```

https://github.com/user-attachments/assets/742ea971-3f23-4d46-a86c-4b34327140b4

m2rads:
I need to just cleanup the test reporter and enhance the test context for AI. I will also cleanup the shortest.config.ts as it is ugly as well. I might actually get rid of it, what do you think? 

As for the location of the test files and having them next to the routes I think I will not get to finish it in this version but I will work on it during the weekend. I am aiming to record a demo so we can launch it on X by EOD.

slavingia:
A lot better! We should remove the console logs imo.

PR #102, opened 2024-12-04
Update readme

m2rads:
- Updated Docs 
- Removed console logs and assert test file


packages/shortest/README.md
```diff
@@ -11,22 +11,23 @@ AI-powered natural language end-to-end testing framework.
...
+pnpm add @antiwork/shortest
 # or
-yarn add shortest
+yarn add @antiwork/shortest
 ```
 
 ## Quick Start
 ```typescript
-import { define } from 'shortest'
+import { UITestBuilder } from '@antiwork/shortest'
```
slavingia, line 23: We should not merge this in. This is so ugly! We should fix this first, then update the README (or do in the same PR as that)

m2rads: Sure.

PR #120, opened 2024-12-19
Update workflow

m2rads:
The Vitest tests are not being maintained and the Vitest workflow keeps failing. 

This PR removes the Vites workflow and enhances Shortest workflow to be more robust in executing the tests. The aim is to bring the branch to a healthy state for better open source collaboration.


m2rads:
@slavingia, @rmarescu 
Maybe we can also remove all the vitests or should we keep them? 

Updating our vitest test and enhancing the web app to have better Vitest test generation is another option we can work on.

slavingia (approved):

Issue #122, opened 2024-12-20
Remove non-shortest tests from repo


slavingia:
### Why

Make it clear what shortest does without polluting the repo with other tests that people may think are shortest tests.

Bounty: $100

martianshaan:
@slavingia  can I work on this ?

slavingia:
Sure, just make sure you keep the shortest tests around, just remove the vitest stuff.

m2rads:
> Apologies for multiple PR creations - was facing an issue while signing the CLA agreement

No problem. Feel free to merge the PR.

PR #126, opened 2024-12-23
Remove non-shortest tests to clarify shortest functionality

khalatevarun:
Fix: #122 

Delete all the tests which are not in the "shortest" scope

Non-shortest test files:
Tests in lib, app, and other directories deal with broader functionalities (dashboard, config, etc.) and don’t directly belong to packages/shortest.


m2rads:
Hi there, please do not delete the shortest tests. They are located in app/__tests__/ directory.

PR #127, opened 2024-12-23
remove non-shortest tests to clarify shortest functionality

khalatevarun:
Fix: https://github.com/anti-work/shortest/issues/122

Delete all the tests which are not in the "shortest" scope

Non-shortest test files:
Tests in lib, app, and other directories deal with broader functionalities (dashboard, config, etc.) and don’t directly belong to packages/shortest.


slavingia:
Looks good, can merge once you sign the CLA. Also, we should add these tests, or similar, using Shortest. Will create a new issue for that.

khalatevarun:
> Looks good, can merge once you sign the CLA. Also, we should add these tests, or similar, using Shortest. Will create a new issue for that.

Created a new [PR](https://github.com/anti-work/shortest/pull/130). I was facing an issue signing the CLA agreement. Thanks!

Issue #132, opened 2024-12-24
Install failure with playwright


kenn:
I was following the install guide but failed at running it.

```sh
% npx shortest
Test Execution Error: browserType.launch: Executable doesn't exist at /Users/kenn/Library/Caches/ms-playwright/chromium-1148/chrome-mac/Chromium.app/Contents/MacOS/Chromium
╔═════════════════════════════════════════════════════════════════════════╗
║ Looks like Playwright Test or Playwright was just installed or updated. ║
║ Please run the following command to download new browsers:              ║
║                                                                         ║
║     npx playwright install                                              ║
║                                                                         ║
║ <3 Playwright Team                                                      ║
╚═════════════════════════════════════════════════════════════════════════╝
⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯

 Tests    0 passed (0)
 Duration   0.07s
 Start at   11:27:15 AM

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
```

Following the instruction, `playwright` seems to be necessary.

```sh
% npx playwright install
sh: playwright: command not found
```

Manually installed `playwright`, then installed browsers manually.

```sh
% pnpm add -D playwright
% npx playwright install
```

So I think you should mention `playwright` somewhere in the install / quick setup section?

m2rads:
Thanks for bringing it up. Will update the docs.

m2rads:
What is the device you're running this on btw?

kenn:
M3 Mac, latest software

kenn:
Also, is there any reason the install guide does not include `--save-dev` or `-D` option?

Not seeing any point in bundling it for production use.

kenn:
Also, add to `.gitignore`

```sh
.shortest
```

m2rads:
Yes valid points. Will address these. Thanks!

m2rads:
[...]

PR #133, opened 2024-12-24
Fix playwright installation issue

m2rads:
Addresses issue #132 

# Changes:
- Fixed installation of playwright browser in setup script
- Add more robust error handling for playwright browser installation - If package manager fails to install the browsers, 
the error handling mechanism will do it. 
- Updated README files to include `-D` in installation guide 
- Update README files to instruct users to add `.shortest/` dir to `.gitignore` file

Version bump: `v0.1.1`


m2rads:
@slavingia fyi the workflow fails because we have a bug in the preview env regarding clerk sign in. Prod and local are stable. I cannot see any issues in Vercel logs either. Need more time to resolve this.

In the meantime, Playwright browser in Vercel and Workflow are being installed properly so this should be a good enough validation for now. 


image

.github/workflows/shortest.yml
```diff
@@ -32,8 +32,8 @@ jobs:
       - name: Install dependencies
         run: pnpm install --no-frozen-lockfile
 
-      - name: Install Playwright browsers
-        run: pnpm exec playwright install
+      - name: Install Playwright
+        run: pnpm exec playwright install chromium
```
slavingia, line 36: Not clear why this is done? Self-review would be helpful!

README.md
```diff
@@ -11,11 +11,16 @@ AI-powered natural language end-to-end testing framework.
...
+npm install -D @antiwork/shortest
 # or
-pnpm add @antiwork/shortest
+pnpm add -D @antiwork/shortest
 # or
-yarn add @antiwork/shortest
+yarn add -D @antiwork/shortest
+```
+
+Add `.shortest/` to your `.gitignore`:
```
slavingia, line 21: why?

.github/workflows/shortest.yml
m2rads: Since we only use chromium browser, I only install this browser to speed up the execution of workflow.

slavingia: Sounds good. I would always include the why behind changes in the PR description/comments, makes it easier to know / approve an approach.

README.md
m2rads: `-D` to add shortest as a dev dependency because it is not needed when bundling an app. 

The other one is just to let the users know that .shortest/ is not needed to be tracked.

m2rads (commented):
So this is the browser manager that manages the state of the browsers for test files. I added a simple exception handling mechanism that would install chromium browsers if they are not found in the users machine. This would prevent the big error message generated by Playwright being thrown in the terminal.

.github/workflows/shortest.yml
m2rads: Sure

README.md
slavingia, line 21:
```suggestion
Add `.shortest/` to your `.gitignore` (where Shortest stores screenshots of each test run):
```

"that .shortest/ is not needed to be tracked" isn't really helpful, that's just explaining what the code does. I think this would be better.

m2rads: Looking at the open PRs right now .shortest/ will soon store more than screenshots. Maybe we can say something more generic like `where shortest stores test related data`!?

slavingia: We can update the readme in those PRs themselves. Each PR should be atomic.

slavingia (approved):
Sure!

Issue #124, opened 2024-12-20
Cache tests


slavingia:
A way to make tests faster.

Via snapshot images? Via URLs?

Bounty: $1,000 for each PR that improves the speed of a shortest test.

bojl:
One of way of speeding up tests is to maintain a cache that maps the test input (prompt, site) to the steps the AI decides to take (mouse move, click, type, button press, etc). That way, during the next identical test run you can just have playwright execute the steps without calling the LLM until the test is complete (to verify that the test was successful). You might also be able to store jest verification steps along the way (like validating divs exist, etc) to avoid the need to call the LLM at all.

This is useful for testing regressions, but obviously not as useful when the page you're testing has changed since the last test run.

Another thing you can do is allow the LLM to suggest multiple action in a single prompt rather than 1 action = 1 prompt

bojl:
Ah I've been beaten to the punch, at least on the first suggestion-- https://github.com/anti-work/shortest/pull/134

slavingia:
Yep, that's probably the right approach. And rerun a step/whole test upon failure in case of a page change.

gladyshcodes:
@slavingia Hi there! That is what I have in mind and how I see caching implemented, what do you think?

Essentially, we could keep track of every step made by Playwright. We could call it action or similarly. Each action will be added to cache. On subsequent test runs, we would replay those instructions. If any of them fail, we will fallback to normal test execution and call LLM. I am also thinking of caching LLM responses. I am not sure about the performance gain and effectivity of such approach yet, just wanted to gather some feedback and thoughts :)

<img width="847" alt="Screenshot 2024-12-26 at 13 35 20" src="https://github.com/user-attachments/assets/2c906082-0059-4d82-827e-09ebf97e1809" />

gladyshcodes:
@slavingia Hi there! That is how I see caching implemented, what do you think?

Essentially, we could keep track of every step made along with LLM resps in cache. I believe it's better to do it per test rather than per test file. On initial run, each step will be stored in cache. On subsequent test runs, we would replay those instructions. If any of them fail or UI have changed, we will fall back to a normal test execution and call LLM, rebuilding cache on the fly. I am not sure about the performance gain and effectiveness of such approach yet, just wanted to gather some feedback and thoughts :)

<img width="1016" alt="Screenshot 2024-12-26 at 15 42 35" src="https://github.com/user-attachments/assets/ed6e929e-8bfe-4254-be46-76b44e86f4a3" />

[...]

PR #134, opened 2024-12-24
add a basic implementation of a recording based tool and client

kshmir:
Some stuff is still quite irrelevant as I was trying out the repo...

1. Made a SnapshotBrowserTool that records the steps taken by the AI.
2. Made a SnapshotAIClient that makes a final decision after all the recorded steps are rerun from the snapshot file (JSONL)

I need to clean it up, would be useful to ignore the screenshot steps, and maybe add a wait for load after clicking or navigating instead of a 1s sleep.

Also a --snapshot parameter should be added and a cleanup parameter as well.

Speed is clearly better, and the AI still makes the final judgement call with a different prompt.

### DIFF
--- a/packages/shortest/src/ai/client.ts
+++ b/packages/shortest/src/ai/client.ts
@@ -10,8 +10,9 @@ export class AIClient {
   private model: string;
   private maxMessages: number;
   private debugMode: boolean;
+  public basePrompt: string;
 
-  constructor(config: AIConfig, debugMode: boolean = false) {
+  constructor(config: AIConfig, debugMode: boolean = true) {
     if (!config.apiKey) {
       throw new Error('Anthropic API key is required. Set it in shortest.config.ts or ANTHROPIC_API_KEY env var');
     }
@@ -21,7 +22,8 @@ export class AIClient {
     });
     this.model = 'claude-3-5-sonnet-20241022';
     this.maxMessages = 10;
-    this.debugMode = debugMode;
+    this.debugMode = true;
+    this.basePrompt = SYSTEM_PROMPT;
   }
 
   async processAction(
@@ -72,7 +74,7 @@ export class AIClient {
           model: this.model,
           max_tokens: 1024,
           messages,
-          system: SYSTEM_PROMPT,
+          system: this.basePrompt,
           tools: [...AITools],
           betas: ["computer-use-2024-10-22"]
         });
--- a/packages/shortest/src/ai/client.ts
+++ b/packages/shortest/src/ai/client.ts
@@ -5,14 +5,24 @@ import { BrowserTool } from '../browser/core/browser-tool';
 import { AITools } from './tools';
 import pc from 'picocolors';
 
-export class AIClient {
+export interface IAIClient {
+  processAction(prompt: string, browserTool: BrowserTool,
+    outputCallback?: (content: Anthropic.Beta.Messages.BetaContentBlockParam) => void,
+    toolOutputCallback?: (name: string, input: any) => void
+  ): Promise<{
+    messages: Anthropic.Beta.Messages.BetaMessageParam[];
+    finalResponse: Anthropic.Beta.Messages.BetaMessage;
+  } | null>;
+}
+
+export class AIClient implements IAIClient {
   private client: Anthropic;
   private model: string;
   private maxMessages: number;
   private debugMode: boolean;
   public basePrompt: string;
 
-  constructor(config: AIConfig, debugMode: boolean = true) {
+  constructor(config: AIConfig, debugMode: boolean = false) {
     if (!config.apiKey) {
       throw new Error('Anthropic API key is required. Set it in shortest.config.ts or ANTHROPIC_API_KEY env var');
     }
@@ -22,7 +32,7 @@ export class AIClient {
     });
     this.model = 'claude-3-5-sonnet-20241022';
     this.maxMessages = 10;
-    this.debugMode = true;
+    this.debugMode = debugMode;
     this.basePrompt = SYSTEM_PROMPT;
   }
 
@@ -46,6 +56,7 @@ export class AIClient {
         await new Promise(r => setTimeout(r, 5000 * attempts));
       }
     }
+    return null;
   }
 
   async makeRequest(
--- a/packages/shortest/src/ai/snapshot-client.ts
+++ b/packages/shortest/src/ai/snapshot-client.ts
@@ -1,6 +1,6 @@
 import { existsSync, readFileSync } from 'fs';
 import { join } from 'path';
-import { AIClient } from './client';
+import { AIClient, IAIClient } from './client';
 import { BrowserTool } from '../browser/core/browser-tool';
 import { ActionInput } from '../types/browser';
 import { TestContext } from '../types/test';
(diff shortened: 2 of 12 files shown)


kshmir:
This aims to implement #124 through snapshotting!

kshmir:
To be 100% fair, thinking it, "recording" is a better name than snapshot in this case. 

https://github.com/vcr/vcr is the main inspiration for me using these types of tests, I got used to snapshots as a "good enough" solution, but in this case recording is the only way.

kshmir:
A key thing will be to recognize where things are being clicked instead of using coordinates, by using XPath, otherwise this approach would be too flaky.

packages/shortest/src/ai/snapshot-client.ts
```diff
@@ -0,0 +1,102 @@
...
+import { join } from 'path';
+import { AIClient } from './client';
+import { BrowserTool } from '../browser/core/browser-tool';
+import { ActionInput } from '../types/browser';
+import { TestContext } from '../types/test';
+import Anthropic from '@anthropic-ai/sdk';
+import { sleep } from '@anthropic-ai/sdk/core';
+import pc from 'picocolors';
+
+const JUDGMENT_PROMPT = `You are a test result validator. Your task is to evaluate if a test passed or failed based on the final state.
```
m2rads, line 11: What is the difference between `JUDGMENT_PROMPT` and `JUGDEMENT_SYSTEM_PROMPT`. Preferably, move this to a new file in `prompts` dir and export from `index.ts`.

packages/shortest/src/ai/client.ts
```diff
@@ -10,8 +10,9 @@ export class AIClient {
   private model: string;
   private maxMessages: number;
   private debugMode: boolean;
+  public basePrompt: string;
 
-  constructor(config: AIConfig, debugMode: boolean = false) {
+  constructor(config: AIConfig, debugMode: boolean = true) {
```
m2rads, line 15: I need to update the docs. We have a CLI arg called `pnpm shortest --debug-ai`. for your `debug.ts` solution please adjust it so that it will show your logs when that CLI arg is provided. Please show your logs along with the already implemented logs that are in place. You can refer to `cli/bin.ts`

packages/shortest/src/browser/core/snapshot-browser-tool.ts
```diff
@@ -0,0 +1,55 @@
...
+    // Only set up recording if snapshot doesn't exist
+    this.shouldRecord = !existsSync(this.snapshotFile);
+    
+    if (this.shouldRecord) {
+      // Create/clear the file only if we're going to record
+      writeFileSync(this.snapshotFile!, '');
+    }
+  }
+
+  async execute(input: ActionInput): Promise<ToolResult> {
```
m2rads, line 36: Seems like this part is executing the test steps twice. Once when AI performs it and once as a replay. Perhaps this should be avoided. Maybe the first time we run `pnpm shortest` we should let AI execute its actions and then run the replay in the consecutive time. Would be cool to have a smart mechanism to delete the snapshots once they become obsolete (No rush for now I will add this as a new issue later - unless you want to tackle this in the same PR).

m2rads (changes requested):
Thanks for your PR @kshmir. A few comments for you to look at :) 

I ran the tests with your new feature and yes the test execution is faster but seems flaky. Seems like the only actions that are properly being executed and recorded are navigation and screenshot actions. If you know a better way to perform browser actions such as xpath instead of coordinates please address them in the same PR otherwise the package will become broken. 

Also I already left you a review about this but I mention it again. Seems like each test step executed by AI is being executed again by the replay mechanism. At least that's what I understood from the debugging logs. Maybe you can help me understand this: 

image
 
Overall your approach seems very interesting and definitely is faster. Please address these issues and let me know if you have any questions or suggestions.

m2rads:
Also please bump the version in `packages/shortest/package.json` to `"version": "0.1.1"` and update the `packages/shortest/CHANGELOG.md` 

Thank you.

slavingia:
I do think snapshots work as a term over recording. Seems more accurate/intuitive to "screenshotting" which is happening, versus recording which implies videos imo.

packages/shortest/src/ai/snapshot-client.ts
kshmir: Cursor just made it twice, fixing soon.

packages/shortest/src/ai/client.ts
kshmir: Yes, will include support for cli arguments to make debugging easier, also for enabling snapshots.

packages/shortest/src/browser/core/snapshot-browser-tool.ts
kshmir: I'll try to make something usable inside a flag to keep moving forward.

kshmir:
> I do think snapshots work as a term over recording. Seems more accurate/intuitive to "screenshotting" which is happening, versus recording which implies videos imo.

Makes sense, will keep it that way then.

kshmir:
> Thanks for your PR @kshmir. A few comments for you to look at :)
> 
> I ran the tests with your new feature and yes the test execution is faster but seems flaky. Seems like the only actions that are properly being executed and recorded are navigation and screenshot actions. If you know a better way to perform browser actions such as xpath instead of coordinates please address them in the same PR otherwise the package will become broken.
> 
> Also I already left you a review about this but I mention it again. Seems like each test step executed by AI is being executed again by the replay mechanism. At least that's what I understood from the debugging logs. Maybe you can help me understand this:
> 
> image
> 
> Overall your approach seems very interesting and definitely is faster. Please address these issues and let me know if you have any questions or suggestions.

image

@m2rads this is what is executed, at the same time, I'll remove the screenshot steps, since they're irrelevant, and convert the click action to a XPath execution.

slavingia:
Why is xpath better than clicking a coordinate? Feels like the latter is more like what a human would do. I can see it breaking more often I guess if the UI changes, but then it would just rerun the test and look for the button like a human would. Seems better to me than falling back on something deterministic like xpath (why not write that explicitly in the test then?) but perhaps I'm missing something.

m2rads:
> Why is xpath better than clicking a coordinate? Feels like the latter is more like what a human would do. I can see it breaking more often I guess if the UI changes, but then it would just rerun the test and look for the button like a human would. Seems better to me than falling back on something deterministic like xpath (why not write that explicitly in the test then?) but perhaps I'm missing something.

I agree with you and I am not convinced that xpath is better option either. Besides in some cases xpath can be flaky. The computer use API is trained on x,y coordinates. Can you explain how you implemented xpath and used it with AI?

kshmir:
@m2rads I'm still testing out the xpath solution... it's obviously a tradeoff...

This is what I see...

image

Some XPaths can be more reliable than others, the most basic is the dom tree structure which is similar to X,Y coordinates, except that inside a DOM. 
data-testid is the "industry standard"
I also asked sonnet to give me a hint in the prompt that I can use to make a text search.

Based on what you say, XPaths could be a fallback, I'm now recording a structure like this

image

kshmir:
I guess for now I'll leave the XPath part in another commits based on what you guys mentioned and release a final version tomorrow that minimizes screenshots. I think we have different goals for what the tool can be doing so I don't want to overdoit.

kshmir:
I think the most important pending thing will be a flag to disable/enable all of this. I let the XPath stuff for other PR.

m2rads:
@kshmir would be nice to resolve the conflicts and then I will test again. Thank you

slavingia:
Reopen if we continue here

Issue #123, opened 2024-12-20
Nested/chained tests


slavingia:
### What

Some way of chaining tests to depend on other tests.

### Why

Save time/lines of code when running tests.

Bounty: $1,000 for a merged PR that solves this problem, $100 for a good proposal.

PR #136, opened 2024-12-24
add chained tests

crabest:
For https://github.com/anti-work/shortest/issues/123, I thought of adding chained tests just like a testA inside testB to let testA depends on testB, but with this approach testA cannot depend on another testC unless testC is between testA and testB.
As a solution we can let the tests seperated but it can depend on multiple tests no matter the position.

Please let me know your thoughts and feedback.


packages/shortest/README.md
```diff
@@ -73,6 +74,39 @@ shortest('Login to the app using username and password', {
...
+  email: 'test@shortest.com', 
+  password: 'password' 
+})
+.expect('User should be logged in successfully');
+
+shortest('Create a new post', { 
+  title: 'Test Post', 
+  content: 'This is a test' 
+})
+.requires('Login to the application')  // This test will run after login
```
bojl, line 93: So something I really like about the library as it exists currently is that it's really elegant and concise, requiring only the `shortest` function call to run a test. IMO adding too many types of chained methods kind of defeats that.

I think it should be possible to have this type of call structure:

```
shortest("test spec")
    .then("dependent test spec 1")
    .then("dependent test spec 2");
``` 
 
What do you think?

crabest: it will still have that problem if want to run it like:
```typescript
shortest("test spec")

shortest("dependent test spec 2");

shortest("dependent test spec 1")
    .requires("test spec", "dependent test spec 2")
```

There is no equivalent way to do it in that approach.

crabest: we can do 

```typescript
shortest("test spec")
    .then("dependent test spec 2")
    .then("dependent test spec 1");
```

but i will force me to let "dependent test spec 2" depend on "test spec" even if I don't want to.

slavingia: Instead of using magic strings, could we assign a variable to the test and then have that be the input, versus a string that matches?

E.g

login = shortest("user can login")

after(login).shortest("user can create post")

or

shortest("user can create post").depends_on(login)

(I prefer the first as more readable.)

crabest: I like the first one too, and there is no constraints when we add multiple dependencies.

after(login, click_settings).shortest("change language to arabic")

crabest:
for less confusion, I named it `afterTest()` that can accept more than one `shortest` variables

slavingia:
I prefer "after"

crabest:
I said for less confusion because we have `after()` method for `shortest` for callback functions
image

what do you think

slavingia:
Hm where's that coming from, if it's not being imported. But good point, I'll think on it

slavingia:
Could we overload it? So you could use both?

crabest:
we can use both, the "less confustion" just for us humans

slavingia:
Let's go with after. I think the simple language is critical for adoption.

crabest:
renamed to "after", now it should be good!

packages/shortest/README.md
```diff
@@ -7,14 +7,20 @@ AI-powered natural language end-to-end testing framework.
 - AI-powered test execution using Claude computer use API
 - Built on Playwright
 - GitHub integration with 2FA support
+- Test dependency chaining with reusable test steps
 
 ## Installation
 ```bash
-npm install @antiwork/shortest
+npm install -D @antiwork/shortest
```
slavingia, line 14: IMO these should be in its own PR, as not related to this feature

packages/shortest/README.md
```diff
@@ -73,6 +79,42 @@ shortest('Login to the app using username and password', {
 });
 ```
 
+## Test Dependencies
```
slavingia, line 82:
```suggestion
## Chained Tests
```

packages/shortest/README.md
```diff
@@ -73,6 +79,42 @@ shortest('Login to the app using username and password', {
...
+You can create reusable test steps and chain them together. This helps organize complex test flows and maintain state between related tests:
```
slavingia, line 83:
```suggestion
Create reusable test steps and chain them together to organize test flows and maintain state between related tests.
```

packages/shortest/README.md
```diff
@@ -73,6 +79,42 @@ shortest('Login to the app using username and password', {
...
+  email: 'test@example.com',
+  password: 'password'
+});
+
+const settings = shortest('Click settings button');
+
+// Chain tests with dependencies
+after(login, settings)
+  .shortest('Change language to Arabic')
+  .expect('Language should be Arabic');
```
slavingia, line 99: Do we need the expects? Feels like something shortest should handle.

packages/shortest/README.md
```diff
@@ -73,6 +79,42 @@ shortest('Login to the app using username and password', {
...
+const changeLanguage = after(login, settings)
+  .shortest('Change language to Arabic');
+
+// Use stored steps in other tests
+after(changeLanguage)
+  .shortest('Verify Arabic UI')
+  .expect('UI elements should be in Arabic');
+```
+
+Benefits:
```
slavingia, line 111: Not useful imo.

slavingia (changes requested):
Commented. Feature itself looks good! I do think we should update this repo with at least one example that can be run locally and in CI.

packages/shortest/README.md
```diff
@@ -73,6 +79,42 @@ shortest('Login to the app using username and password', {
...
+const login = shortest('User can login', { 
+  email: 'test@example.com',
+  password: 'password'
+});
+
+const settings = shortest('Click settings button');
+
+// Chain tests with dependencies
+after(login, settings)
+  .shortest('Change language to Arabic')
```
slavingia, line 98:
```suggestion
  .shortest('Language can be changed to Arabic')
```

packages/shortest/README.md
```diff
@@ -73,6 +79,42 @@ shortest('Login to the app using username and password', {
...
+
+// Store intermediate steps
+const changeLanguage = after(login, settings)
+  .shortest('Change language to Arabic');
+
+// Use stored steps in other tests
+after(changeLanguage)
+  .shortest('Verify Arabic UI')
+  .expect('UI elements should be in Arabic');
+```
```
slavingia, line 109:
```suggestion
// Use stored steps in other tests
shortest(login).shortest(changeLanguage).shortest(changeLanguageBack)
```

What about something like this? Keeps it all as shortest and keeps it very clean and simple.

slavingia: This may mean that writing tests could be as simple as:

```
shortest(login).shortest(changeLanguage).shortest(changeLanguageBack)
```

```
shortest(login).shortest(changeLanguageToJapanese).shortest(changeLanguageBackToEnglish)
```

And AI could automatically "cache" the login step?

(This gets a little crazy so maybe in another PR. haha)

packages/shortest/README.md
crabest: if there is no expectations we don't add them to the prompt, but we can specify it in the shortest string. it's just a preference.
image

packages/shortest/README.md
crabest: instead of expect the developer can use this approach

packages/shortest/README.md
crabest: we can remove it since it's a dev docs.

packages/shortest/README.md
crabest: man, love that, it sounds great, but I agree with the last! let's make it seperated that this

crabest: this approach is great but I don't think it will replace this approach of `after()` function since we can depend on one shortest test in mutliple `after()`

```typescript
const login = shortest("login ...", {...credentials})

after(login).shortest("The dark mode could be changed")
after(login).shortest("The language should be english")
```

crabest: ofc we can develop it in that approach too if we checked the type of the input if it's ChainTest or string.

slavingia: Yeah think we should detect!

slavingia:
ideas from internal discussion:

```
const login = shortest('user can login with email and password')
shortest(login).shortest('user can modify their account-level refund policy from the default of 30-days to no refunds allowed')

const loginAsLawyer = shortest('...')
const loginAsContractor = shortest('..')
shortest(loginAsLawyer.allAppActions)
shortest(loginAsContractor.allAppActions)
const allAppActions = [shortest('send invoice to company')];
```

This would make it even shorter.

crabest:
let me in those discussions (:
for the array part, this would be shorter and more readable.

```typescript
const allAppActions = shortest(['send invoice to company', 'view invoices'])
```

crabest:
10am EST Friday, sadly I would still be at work.
for those changes, let's iterate in the same pull request? what do you think

slavingia:
yep sounds good!

crabest:
with those changes, I think let's drop that `after()`

crabest:
using your example, this is how it will look like now:

```typescript
const login = 'user can login with email and password'
shortest([login, 'user can modify their account-level refund policy from the default of 30-days to no refunds allowed'])

const loginAsLawyer = '...'
const loginAsContractor = '...'
const allAppActions = ['send invoice to company', 'view invoices'];
shortest([loginAsLawyer, ...allAppActions])
shortest([loginAsContractor, ...allAppActions])
```

slavingia:
Neato! @rmarescu @m2rads et al, happy w/ that?

crabest:
my vision is to be like that:

```typescript
const login = shortest('user can login with email and password')
shortest(login).shortest('user can modify their account-level refund policy from the default of 30-days to no refunds allowed')

const loginAsLawyer = shortest('...')
const loginAsContractor = shortest('...')
const allAppActions = ['send invoice to company', 'view invoices']
shortest(loginAsLawyer).shortest(allAppActions)
shortest(loginAsContractor).shortest(allAppActions)
```

let me know what do you think

slavingia:
I like the shorter one, without the many "shortest"s personally. I was convinced! I can send you the recording of the call if you wish

crabest:
i would love to, how

slavingia:
DM me on X or email me sahil.lavingia@gmail.com and will send!

crabest:
DMed on X

m2rads:
Seems like your new changes are not resolving properly. 

image

### Here's how you can fix it: 
When making changes to `src/index.ts` you'd need to update `index.d.ts` as well.  
We have been dealing with some issues when generating `index.d.ts` when bundling the package so we implement it manually. Please make sure your new changes are reflected in that file as well. 

### Here's why this issue happens: 

`src/index.ts` is our source file that contains the actual implementation and gets compiled to JavaScript.

`index.d.ts` is our declaration file that only contains type definitions and Is used by TypeScript for type checking. It also defines the public API types for consumers.

We need both because the compiled JavaScript `(dist/index.js)` doesn't contain type information and external users of our package need type definitions for TypeScript support.

I will open a new issue and explain how we can fix this. Please let me know if you have any questions.

m2rads:
Would be good to have a demo using this new feature when you have finished your implementation.

crabest:
> ```ts
> const login = 'user can login with email and password'
> shortest([login, 'user can modify their account-level refund policy from the default of 30-days to no refunds allowed'])
> 
> const loginAsLawyer = '...'
> const loginAsContractor = '...'
> const allAppActions = ['send invoice to company', 'view invoices'];
> shortest([loginAsLawyer, ...allAppActions])
> shortest([loginAsContractor, ...allAppActions])
> ```

the example you used is not implemented, only the example above ^

m2rads:
Looks good to me. Would be nice to write some new tests using the new chained tests. 

To prepare for release please update your branch and bump the version to `0.3.1`. I think this version is appropriate as this PR is a major feature. Also please update changelogs to add your changes.

slavingia:
0.3.0 imo

m2rads:
Can we export a test and use it in another folder?

m2rads:
> Looks good to me. Would be nice to write some new tests using the new chained tests.
> 
> To prepare for release please update your branch and bump the version to `0.3.1`. I think this version is appropriate as this PR is a major feature. Also please update changelogs to add your changes.

@crabest You don't need to do this anymore. Once review is done will just merge to main.

slavingia:
Please merge main to include a new caching feature!

crabest:
Done!

README.md
```diff
@@ -130,6 +130,30 @@ shortest.afterAll(async ({ page }) => {
 });
 ```
 
+### Chained Tests
```
slavingia, line 133:
```suggestion
### Chaining tests
```

slavingia (approved):
Lgtm

PedroAVJ:
Great work guys! Was just looking to do a pattern like this for longer test workflows

slavingia:
Made one change: https://github.com/anti-work/shortest/pull/136/commits/f7dc8df0aad5c68308c629f6f031423cadf5b8a0

IMO, we should try to avoid "expect" as that's a pre-AI approach to writing tests. The AI can determine this and fail the test of any of the other test conditions (in natural language) failed.

slavingia:
If you agree, please merge or let me know and I can do so!

crabest:
sure, we just need to adjust the the prompt to let it understand to get the expectation the magic string if we didn't pass those exceptions.

Issue #145, opened 2024-12-26
Linter


slavingia:
Use prettier, eslint, or biome. 

Why: Save time reviewing PRs, less opinions to think about.

PR #146, opened 2024-12-26
configure eslint and run it on project to fix auto fixable error & warnings

khalatevarun:
Fix: #145


m2rads:
We might have an issue in the preview env that prevents from our workflow to pass. Please allow me to solve that issue so we can run our e2e test on this PR to Validate.

Issue #144, opened 2024-12-25
playwright browser installation does not consider the package manger used


amk-dev:
**How to reproduce**

1. make sure to remove playwright cache, even if you uninstall playwright, the cache may still be there
   - in mac its `~/Library/Caches/ms-playwright`
   - for other platforms check playwright the [docs](https://playwright.dev/docs/browsers#managing-browser-binaries)
2. make a new directory `shortest-test`, cd into it, init & install `@antiwork/shortest` using `pnpm`
3. create `shortest.config.ts`, here's mine
```typescript
import type { ShortestConfig } from "@antiwork/shortest";

export default {
  headless: false,
  baseUrl: "http://localhost:3000",
  testDir: "app/__tests__",
  anthropicKey: "sample-key",
} satisfies ShortestConfig;
```
4. create a sample test file `app/__tests__/helloworld.test.ts`, doesnt matter if its empty
5. run `pnpm shortest`
6. you'll get the below error screen

<img width="1444" alt="image" src="https://github.com/user-attachments/assets/3727c9df-9547-48ca-aa13-fd7eff885f06" />

I can make a PR with a fix once you guys confirms the bug. lmk.

m2rads:
Thanks for reporting this issue. Could you please run npx playwright install in the same project directory and let me know if you see any issues!? 

If not then we can improve this by handling each package manager separately instead of using npx for all. Please let me know. Thanks

slavingia:
Reopen once ^

amk-dev:
hey this still happens. i'm guessing `npx` is being confused by the `node_modules` structure used by `pnpm` and giving up.

<img width="1448" alt="image" src="https://github.com/user-attachments/assets/ff0a283c-af03-41f8-b3e7-1db1959c0acc" />

---

i was exploring the codebase, so gave solving this a shot, the below solution should work. ( i tested pnpm on mac with it. incase we're going with this solution, i can do further testing and make a PR ).

Also when going through the code, saw the browser installation logic used in multiple places,

- [cli/setup.ts](https://github.com/anti-work/shortest/blob/28f02bdd7886940b77d6ab72d88d010ddbce7d88/packages/shortest/src/cli/setup.ts#L23-L32)
- [browser/manager/index.ts](https://github.com/anti-work/shortest/blob/28f02bdd7886940b77d6ab72d88d010ddbce7d88/packages/shortest/src/browser/manager/index.ts#L31-L48)

[...]

PR #169, opened 2024-12-26
fix: Improve postinstall script compatibility for Windows

cutsin:


m2rads:
Hi @cutsin thanks for the PR. We're thinking that this postinstall script is obsolete now because when the package is finished installing the setup scripts is still not available to be executed. We are tracking  this in here: issue #144

cutsin:
Thanks, understood. Would you consider merging this as a temp fix before #144?
I'll keep following up on subsequent issues.

PR #173, opened 2024-12-26
fix: consider package managers the user uses when installing browsers via playwright

amk-dev:
Fixs #144 

**Changes**

1.  consider the package manager the user is using when installing chromium with playwright.
2. the current package manager & the execute command is determinded using [antfu-collective/package-manager-detector](https://github.com/antfu-collective/package-manager-detector)
3. removes the `setup.ts` as discussed [here](https://github.com/anti-work/shortest/issues/144#issuecomment-2562045441)
4. cleans up the `package.json` scripts that uses `setup.ts`


packages/shortest/package.json
```diff
@@ -25,11 +25,11 @@
     "build": "rm -rf dist && pnpm build:types && pnpm build:js && pnpm build:cli",
     "prepare": "pnpm build",
     "prepublishOnly": "pnpm build",
-    "postinstall": "node -e \"try { if (require('fs').existsSync('./dist/cli/setup.js')) require('./dist/cli/setup.js') } catch (e) {} \" && (chmod +x dist/cli/bin.js || true)",
```
shwetasngh, line 28: "postinstall": "node -e \"try { if (require('fs').existsSync('./dist/cli/setup.js')) require('./dist/cli/setup.js'); } catch (e) {}\" && (node -e \"if (process.platform !== 'win32') { try { require('child_process').execSync('chmod +x dist/cli/bin.js'); } catch (_) {} }\")"

amk-dev: hey, i do not understand what you meant here. this PR removed the `setup.ts`. so the above is no longer needed.

hence resolving the conversation.

m2rads:
Thanks for waiting will review this soon.

slavingia (approved):

slavingia:
Merge conflicts to resolve!

slavingia:
Failed due to not having Anthropic key, not sure why

amk-dev:
> Failed due to not having Anthropic key, not sure why

i've tested locally and no issues found. might be some config issue ?

slavingia:
They'll be published, right? Just not included in version notes.

amk-dev:
> They'll be published, right? Just not included in version notes.

adding my  two cents

usually libraries include even tiny changes,

in this example, having this might not hurt, and even might be helpful when reading back after sometime.

bugfixs

- consider package managers the user uses when installing browsers via playwright (#PRNUMBER)
- remove duplicate line from system prompt (#PRNUMBER)

if you're communicating to users through social channels, like twitter / your blog, then it makes sense to skip these and focus on the major things.

but in github, since its mostly devs reading, everything is good to include imo.

PR #177, opened 2024-12-26
Add husky and lint-staged to run pre commit scripts like eslint on committed files

khalatevarun:
Fix: #178


slavingia:
Prefer autofix

PR #179, opened 2024-12-26
Add caching

gladyshcodes:
**Issue #124** 

This PR introduces basic caching mechanism to reduce costs and increase effectiveness of running test suites.

**Performance Boost:** Achieves an average speedup of 400%-600%, automations like "Find Lionel Messi Wikipedia page" now completing up to 5x faster

**Flow:**
Screenshot 2024-12-27 at 00 41 23

**Note:** This PR is WIP and still on early stage. It's there early to gather feedback. Future tests, refinements and enhancements will be made to make it reliable and even more performant

### DIFF
--- a/packages/shortest/src/ai/client.ts
+++ b/packages/shortest/src/ai/client.ts
@@ -4,6 +4,9 @@ import { SYSTEM_PROMPT } from './prompts';
 import { BrowserTool } from '../browser/core/browser-tool';
 import { AITools } from './tools';
 import pc from 'picocolors';
+import { BaseCache } from '../cache/cache';
+import { CacheAction, CacheEntry, CacheStep } from '../types/cache';
+import { TestFunction } from '../types';
 
 export class AIClient {
   private client: Anthropic;
@@ -13,11 +16,13 @@ export class AIClient {
 
   constructor(config: AIConfig, debugMode: boolean = false) {
     if (!config.apiKey) {
-      throw new Error('Anthropic API key is required. Set it in shortest.config.ts or ANTHROPIC_API_KEY env var');
+      throw new Error(
+        'Anthropic API key is required. Set it in shortest.config.ts or ANTHROPIC_API_KEY env var'
+      );
     }
 
     this.client = new Anthropic({
-      apiKey: config.apiKey
+      apiKey: config.apiKey,
     });
     this.model = 'claude-3-5-sonnet-20241022';
     this.maxMessages = 10;
@@ -27,7 +32,11 @@ export class AIClient {
   async processAction(
     prompt: string,
     browserTool: BrowserTool,
-    outputCallback?: (content: Anthropic.Beta.Messages.BetaContentBlockParam) => void,
+    cache: BaseCache<CacheEntry>,
+    test: TestFunction,
+    outputCallback?: (
+      content: Anthropic.Beta.Messages.BetaContentBlockParam
+    ) => void,
     toolOutputCallback?: (name: string, input: any) => void
   ) {
     const maxRetries = 3;
@@ -35,13 +44,20 @@ export class AIClient {
 
     while (attempts < maxRetries) {
       try {
-        return await this.makeRequest(prompt, browserTool, outputCallback, toolOutputCallback);
+        return await this.makeRequest(
+          prompt,
+          browserTool,
+          cache,
+          test,
+          outputCallback,
+          toolOutputCallback
+        );
       } catch (error: any) {
         attempts++;
         if (attempts === maxRetries) throw error;
-        
+
         console.log(`Retry attempt ${attempts}/${maxRetries}`);
-        await new Promise(r => setTimeout(r, 5000 * attempts));
+        await new Promise((r) => setTimeout(r, 5000 * attempts));
       }
     }
   }
@@ -49,7 +65,11 @@ export class AIClient {
   async makeRequest(
     prompt: string,
     browserTool: BrowserTool,
-    outputCallback?: (content: Anthropic.Beta.Messages.BetaContentBlockParam) => void,
+    cache: BaseCache<CacheEntry>,
+    test: TestFunction,
+    outputCallback?: (
+      content: Anthropic.Beta.Messages.BetaContentBlockParam
+    ) => void,
     toolOutputCallback?: (name: string, input: any) => void
   ) {
     const messages: Anthropic.Beta.Messages.BetaMessageParam[] = [];
@@ -61,12 +81,12 @@ export class AIClient {
 
     messages.push({
       role: 'user',
-      content: prompt
+      content: prompt,
     });
 
     while (true) {
       try {
-        await new Promise(resolve => setTimeout(resolve, 1000));
+        await new Promise((resolve) => setTimeout(resolve, 1000));
 
         const response = await this.client.beta.messages.create({
           model: this.model,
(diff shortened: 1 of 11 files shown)


gladyshcodes:
```
async getComponentStringByCoords(x: number, y: number) {
    return await this.getPage().evaluate(
      ([x, y]) => {
        const elem = document.elementFromPoint(x, y);
        return elem?.outerHTML.trim().replace(/\s+/g, ' ');
      },
      [x, y]
    );
  }
```
The function above retrieves normalized component string given X and Y coordinates. The reason it's implemented this way is that currently, Playwright does not support selecting DOM elements using coordinates (see [this closed issue](https://github.com/microsoft/playwright/issues/13273)). The only way I have found was [to use native document func](https://developer.mozilla.org/en-US/docs/Web/API/Document/elementFromPoint). If you have any other ideas in mind, please suggest 👍

m2rads:
Would be nice to resolve the conflicts and then I can test your branch locally.

gladyshcodes:
> Would be nice to resolve the conflicts and then I can test your branch locally.

@m2rads I have resolved conflicts now. You can check it out.

gladyshcodes:
Refinements / Improvements:

1. Write new text execution to cache file only after it has been successfully completed. That way we can be sure cached tests are only successful ones
2. Delete cache entry if it fails to run (e.g UI change) and rewrite it

m2rads:
@gladyshcodes sorry we made some more changes. Please resolve the conflicts again and I will test soon. Thank you :)

gladyshcodes:
> @gladyshcodes sorry we made some more changes. Please resolve the conflicts again and I will test soon. Thank you :)

Hi there @m2rads. Those conflicts actually do not interfere with functionality added. I resolved them 👍 .

packages/shortest/src/types/browser.ts
```diff
@@ -13,23 +13,26 @@ export interface BrowserToolInterface {
...
+  ClearSession = "clear_session",
+  Type = "type",
+  Key = "key",
+  RunCallback = "run_callback",
+  Navigate = "navigate",
+  Sleep = "sleep",
+  CheckMail = "check_mail",
+}
+
+export type BrowserAction = `${BrowserActionEnum}`;
```
m2rads, line 35: Why this change?

m2rads:
Seems like Vercel deployment is failing. I am also getting some build errors locally. 

Steps to reproduce this error: 

Remove these files: 
`pnpm-lock.json`
`node_modules`
`packages/shortest/node_modules`
`packages/shortest/dist`

Then run `pnpm install`

image

packages/shortest/src/types/browser.ts
gladyshcodes: I needed to use `screenshot` browser action and founded it no good to make hard code comparison. That's why I created enum from interface, not touching the interface itself. Now it possible to do like: `BrowserActionEnum.Screenshot`

gladyshcodes:
@m2rads Build issue resolved, but Vercel pipeline does not re-run. I guess your approval is needed

packages/shortest/src/cache/cache.ts
```diff
@@ -0,0 +1,219 @@
...
+  // Locking (ensures that only one process or thread can access or modify cache)
+  private lockFile: string;
+  private readonly LOCK_TIMEOUT_MS = 1_000;
+  protected lockAcquired = false;
+  protected lockAcquireFailures = 0;
+
+  constructor() {
+    this.logger = new Logger();
+    this.cacheFile = path.join(process.cwd(), ".cache", "cache.json");
+    this.lockFile = path.join(process.cwd(), ".cache", "cache.lock");
```
m2rads, line 25: I think it would be better to put this inside `.shortest` folder. Also does this only cache passing test cases?

gladyshcodes: > I think it would be better to put this inside .shortest folder.

Good point. I will rename it

> Also does this only cache passing test cases?

Not entirely sure what you mean. If you are asking whether test file only contains tests that succeeded, then yes

m2rads:
Thanks the build issue is resolved now but I think there might be another issue. When running a test, it caches the steps successfully, but on a consecutive run, it deletes the cache and does the same process. It doesn't seem like the consecutive test is being executed from the cache as it takes the same amount of time to run.


https://github.com/user-attachments/assets/ba574583-5194-4827-8bb6-8550e636d3d6

gladyshcodes:
Hmm that's weird. I have just tried running several tests it and it worked fine for me. I didn't try it with new `dashboard` test though. I will see now and let you know
```shortest("Open Google");```

Screenshot 2024-12-29 at 23 19 05

**Comment by m2rads** [comment]
I see. The new test needs setup with Mailosaur. Maybe try a more complicated test that has multiple steps and see if this issue happens.

**Comment by gladyshcodes** [comment]
@m2rads I have tried with this test, it works:

`shortest("Open first article on Hackernews");`

Screenshot 2024-12-29 at 23 33 15
Screenshot 2024-12-29 at 23 33 22

**Comment by m2rads** [comment]
> @m2rads I have tried with this test, it works:
> 
> `shortest("Open first article on Hackernews");`
> 
> Screenshot 2024-12-29 at 23 33 15 Screenshot 2024-12-29 at 23 33 22

Hhhm seems like this happens when you have multiple test cases. Can you try with more than one test case?

**Comment by gladyshcodes** [comment]
@m2rads Thank you for caching the bug! I've realized the problem was caused by several factors:
1. Missing sleep time after each `browserTool.execute`. Similar to how it's done in runner, we need to address this. However, I believe we should eliminate such code altogether since it affects the performance of both cached and uncached tests. Sometimes, waiting less than a second is sufficient for the next action, while other times, potentially, a longer wait might be needed
2. For `dashboard.test.ts`:
- I managed to get it running by referencing to the source code of #183, as the relevant documentation appears to be missing. It would be helpful under the [services configuration section](https://github.com/anti-work/shortest?tab=readme-ov-file#services-configuration) section. Also, Clerk configuration needs updating: you’ll need to disable the “Require the same device and browser” checkmark in the Clerk dashboard for the tests with mailosaur to work (at least in my case)
- Issue itself was the difference in <a> components from email letter as their `href` attrs differ every time. We now recursively clean up `href` attributes from the tags. The idea that if the URL changes and we need to click the link, the next step will fail

I ran several tests and they pass now:

Screenshot 2024-12-30 at 17 13 21

**Perf boost:** 5.6x

@slavingia If you have time, take a look as well, maybe any suggestions?

File: packages/shortest/src/ai/client.ts [hunk 3]
```diff
@@ -78,6 +82,7 @@ export class AIClient {
     while (true) {
       try {
         await new Promise((resolve) => setTimeout(resolve, 1000));
+        await new Promise((resolve) => setTimeout(resolve, 1000));
```
slavingia, line 85: Is this on purpose?

slavingia:
Merged in auto fix which led to some new failures

slavingia (commented):
Minor comments, looks great otherwise!

packages/shortest/src/ai/client.ts
gladyshcodes: No, it's not. Removed

slavingia (approved):

m2rads (approved):

slavingia:
Bug report:

> however I've removed the button I was testing and it still pass

> maybe I did something wrong, my test is just this:
shortest("Open the home page and signin via email").expect("A message to check your email");

> what I did now is that I removed the "Signing via Email" link and it still pass

> it seems the second time I run the tests they are run against the screenshots but don't hit the server

You can see the video here: https://x.com/madarco/status/1874141497404363128

I guess there needs to be some logic to destroy the cache, so that the spec starts to fail. We'll need to think more on the best way to do that dynamically; we can start by making it very easy (one command) to nuke the cache?

Issue #191, opened 2024-12-27
bug: bundled types are not resolving properly


amk-dev:
**How to reproduce**
1. Get a fresh installation of shortest & 
2. See if in the `shortest.config.ts`, `ShortestConfig` resolves properly

https://github.com/user-attachments/assets/231ce9bb-ef88-4510-91f7-996c3e5d7fd6

amk-dev:
working on a fix fox this.

do you guys care about common js support, or just esm works ? ( ESM only would make the setup + everything very simple imo )

i'm using `are-types-wrong` to evalute the types 

[https://arethetypeswrong.github.io/?p=%40antiwork%2Fshortest%400.1.1](https://arethetypeswrong.github.io/?p=%40antiwork%2Fshortest%400.1.1
)

m2rads:
We have room to improve types when bundling the but are they failing rn in your feature branch? Can you show me the error message you're getting.

m2rads:
There is definitely room for improving the types when bundling. One challenge we ran into early on was generating the `index.d.ts` file with typescript. Right now we are doing it in a hacky way by implementing that file manually. I think the fix would be to generate type along with the build command.

amk-dev:
this is not from a feature branch.
this is from `@antiwork/shortest`@0.1.1 published on npm.

<img width="1226" alt="image" src="https://github.com/user-attachments/assets/fd69841c-8ed3-475d-865a-0ebb37a9952e" />

as you can see the `ShortestConfig` is not resolving

amk-dev:
> One challenge we ran into early on was generating the index.d.ts file with typescript. Right now we are doing it in a hacky way by implementing that file manually.

What should be the public api, knowing that should help me fix it 

but if you want a quick fix without going there, updating the path as mentioned above should work, since the `ShortestConfig` is an important type imo.

m2rads:
So the public API is this: `index.ts` in the root of `packages/shortest` dir. `index.d.ts` should be created automatically by typescript automatically but we are actually writing it manually. So I think that is the main issue we are dealing with. I am open to ideas on how to fix it or if you think we need to address this issue from somewhere else let me know.

PR #193, opened 2024-12-27
fix: fix incorrect path for types in index.d.ts

amk-dev:
Fixs #191 

**Changes**

Previously when the package is bundled, `index.d.ts` had incorrect paths for some imports. this makes the `ShortestConfig` from `@antiwork/shortest` not resolve properly.

here's the file structure of the published package ( i've removed irrelevent files from the tree ) 

```
@antiwork/shortest/
├── dist/
│   └── types/
│       └── types/
│           ├── config.d.ts
│           └── test.d.ts
└── index.d.ts
```

so the imports in `index.d.ts` had two wrong imports
* `./dist/types/test`
* `./dist/types/config`

both will not resolve, the correct paths are 
* `./dist/types/types/test`
* `./dist/types/types/config`

this PR makes that update.

note: as discussed in the issue, there are room for improvement in the generation of `index.d.ts`. I'm gathering context for that. but since this is a rather important fix (imo). raising a PR.

### DIFF
--- a/packages/shortest/index.d.ts
+++ b/packages/shortest/index.d.ts
@@ -1,8 +1,8 @@
 import type { Expect } from "expect";
 import type { Page, Browser, APIRequest, APIRequestContext } from "playwright";
 import type * as playwright from "playwright";
-import type { ShortestConfig } from "./dist/types/config";
-import type { TestAPI, TestContext } from "./dist/types/test";
+import type { ShortestConfig } from "./dist/types/types/config";
+import type { TestAPI, TestContext } from "./dist/types/types/test";
 
 declare global {
   const expect: Expect;


m2rads:
@amk-dev thanks for the PR. Once you have the conflicts resolved I will test your branch and let you know.

amk-dev:
just rebased.

amk-dev:
since the type for `ShortestConfig` is in effect, there are some type errors in `shortest/shortest.config.ts`, that's causing the CI to fail i think.

image

how do you want to handle it ?

@m2rads

slavingia:
Why do we have a types folder within a types folder?

amk-dev:
> Why do we have a types folder within a types folder?

the `build:types` command is building types for `src` into `dist/types`, `src` has another folder named `types`. so `types/types` becomes a thing.

instead of updating the `index.d.ts` like i did in this PR, may be we can just build the `types` into  `dist` instead of `dist/types`, that would eliminate the double types path. but need to see if it breaks any other imports.

image

m2rads:
I think this error is happening because in `types/config.ts` we are enforcing the `anthropicApiKey` to be string but the type for `process.env.ANTHROPIC_API_KEY ` is `string | undefined` therefore typescript is now correctly identifying the type issue here. IMO we could just directly use `process.env.ANTHROPIC_API_KEY ` inside `client.ts` file where our AI implementation exists. This would remove the need for end users to specify their API key in the config file. Sort of like nextjs AI sdk where it reads the API keys internally.  Same with Mailosaur. @slavingia what do you think? If not then we can just update the types to be  `string | undefined` or instruct ts to not be strict about this somehow. 

image


> since the type for `ShortestConfig` is in effect, there are some type errors in `shortest/shortest.config.ts`, that's causing the CI to fail i think.
> 
> image
> how do you want to handle it ?
> 
> @m2rads

slavingia:
Makes sense to do for Anthropic API key. Mailosaur isn't causing an issue as optional so imo better to expose in the config.

amk-dev:
this as accidently closed when i removed the forked repo from my account not remembering this PR was open. let me know if i should reopen this. but if there are plans for #204 to auto generate `index.d.ts`, this PR is not needed.

I've shared some ideas on the same on #204. not making a PR for that issue, because it was assigned to @m2rads.

PR #194, opened 2024-12-27
refactor(config): update config to detect test.ts files instead of files under test dir

khalatevarun:
Fix: #190


slavingia:
Lint and shortest tests are failing

slavingia:
May also want to version this as a breaking change. Can be minor as small and early

packages/shortest/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@antiwork/shortest",
-  "version": "0.2.1",
+  "version": "0.3.0",
```
slavingia, line 3:
```suggestion
  "version": "0.2.2",
```

m2rads:
Shortest workflow is failing with this error. I am getting a similar issue locally. Another issue that came up is when deleting node_modules and building and installing the package again, some of the packages are not resolving properly. Shown in the last image below. 

image

image

slavingia:
^ our fault? I still think we should remove pnpm and go to npm.

If someone else could replicate that and fix, would be great. Don't think anything in this branch would cause it, and someone else reported that there may be inconsistencies in the node modules based on npx vs using pnpm in places.

slavingia:
How does this work if there's a set of tests I don't want to run, or a folder I only want to run tests for, temporarily (via CLI)?

amk-dev:
image



I can see a call to `getConfig` here in `bin.ts`, before the initialize is called.


and `getConfig` throws if the globalConfig is not present.

image


but anything changed recently on this logic ?

packages/shortest/src/cli/bin.ts
```diff
@@ -128,10 +129,13 @@ async function main() {
   const targetUrl = args
     .find((arg) => arg.startsWith("--target="))
     ?.split("=")[1];
-  const testPattern = args.find((arg) => !arg.startsWith("--"));
+  const cliTestPattern = args.find((arg) => !arg.startsWith("--"));
   const debugAI = args.includes("--debug-ai");
 
   try {
+    const config = getConfig();
```
amk-dev, line 136: seems like this code is giving the config error. you're calling `getConfig` before a call to initialize.

amk-dev:
> Another issue that came up is when deleting node_modules and building and installing the package again, some of the packages are not resolving properly. Shown in the last image below.

I tried reproducing this, but no luck.  here are the steps i took,
1. removed `node_modules` from `packages/shortest` & root `node_modules`
2. `pnpm i` at the workspace root

no issue with the imports. may be you can try reloading vscode and see it it fixs the issue. ( `Cmd+Shift+P -> Developer: Relaod Window` ), sometime it happens.

m2rads:
Just posting it here again in case you missed it: 

You're right. with the new implementation we should first initialize the runner and then initialize the config: This should fix it: 


```typescript

  try {
    const runner = new TestRunner(
      process.cwd(),
      true,
      headless,
      targetUrl,
      debugAI
    );
    await runner.initialize();

    const config = getConfig();
    const testPattern = cliTestPattern || config.testPattern;

    await runner.runTests(testPattern);
  }
```

Let me know if you need the full code for this. 

> > Another issue that came up is when deleting node_modules and building and installing the package again, some of the packages are not resolving properly. Shown in the last image below.
> 
> I tried reproducing this, but no luck. here are the steps i took,
> 
> 1. removed `node_modules` from `packages/shortest` & root `node_modules`
> 2. `pnpm i` at the workspace root
> 
> no issue with the imports. may be you can try reloading vscode and see it it fixs the issue. ( `Cmd+Shift+P -> Developer: Relaod Window` ), sometimes it happens.

slavingia:
Please merge main to include a new caching feature!

slavingia:
Would be great to merge this soon so I can use this new approach :)

slavingia:
Would be great to get to this soon. If it's not done by Friday we'll have some internally do it as I think this is quite important!

m2rads:
> Would be great to get to this soon. If it's not done by Friday we'll have some internally do it as I think this is quite important!

Updated the branch and resolved the issue.

packages/shortest/src/cli/bin.ts
```diff
@@ -146,12 +147,9 @@ async function main() {
...
     await runner.initialize();
-
-    if (testPattern) {
-      await runner.runFile(testPattern);
-    } else {
-      await runner.runAll();
-    }
+    const config = getConfig();
+    const testPattern = cliTestPattern || config.testPattern;
+    await runner.runTests(testPattern);
```
m2rads, line 152: Initializing the runner before config file

README.md
```diff
@@ -48,7 +48,7 @@ yarn shortest   # for yarn
   export default {
     headless: false,
     baseUrl: 'http://localhost:3000',
-    testDir: 'app/__tests__',
+    testPattern: '**/*.test.ts',
```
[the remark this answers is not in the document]
slavingia: I think it should be optional and default to what we have if it isn't there. Can do as its own atomic PR.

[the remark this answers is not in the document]
slavingia: I don't think that approach will work, since there may be tests you don't want shortest to run. Like if you're migrating from one framework to another

PR #195, opened 2024-12-28
Update pnpm-lock.yaml

rmarescu:
Diff generated after running `pnpm install` with `pnpm` version [`9.7.1`](https://github.com/anti-work/shortest/blob/08a763477efe26b6b7fcd2d4523030edbd32d230/package.json#L4)

### DIFF
--- a/.github/workflows/autofix.yml
+++ b/.github/workflows/autofix.yml
@@ -15,7 +15,7 @@ jobs:
 
       - uses: pnpm/action-setup@v2
         with:
-          version: 8
+          version: 9.7.1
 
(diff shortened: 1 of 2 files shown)


.github/workflows/autofix.yml
```diff
@@ -15,7 +15,7 @@ jobs:
 
       - uses: pnpm/action-setup@v2
         with:
-          version: 8
+          version: 9.7.1
```
rmarescu, line 18: Should match https://github.com/anti-work/shortest/blob/08a763477efe26b6b7fcd2d4523030edbd32d230/package.json#L4

PR #204, opened 2024-12-29
Remove duplicate line from system prompt

elie222:


m2rads:
Please bump the version in `packages/shortest/package.json` to `v0.2.2`. 

You'd also need to update the `Changelogs` to include your changes.

slavingia:
Don't think this requires a version bump

m2rads:
> Don't think this requires a version bump

Ok in that case we should remember that the next version bump will include this PR and PR: #173

Issue #168, opened 2024-12-26
Release branch


slavingia:
Why:

- Common
- Helps with tracking releases

slavingia:
## [0.1.1] and earlier - 2024-11-28 to 2024-12-24

### Initial Release Features
- AI-powered test execution using Claude 3.5 Sonnet
- Natural language test writing support
- GitHub integration with 2FA support and improved error handling
- Browser automation using Playwright with page object model support
- CLI tool for running tests with headless mode
- Support for ESM and CommonJS
- Mouse tracking and click animations
- Lifecycle hooks including .after() method
- Automatic retry and error handling
- Browser navigation and cleanup improvements
- Flexible test configuration and reporting
- Screenshot management with auto-cleanup
- Windows support for Playwright installation
- TypeScript integration and type declarations
- Environment variable support via dotenv

slavingia:
Any luck?

m2rads:
Alright so here is the same release strategy that is followed by top open source project like Next.js but a bit simpler. 

- The developers continue working on their features as usual by creating a feature branch and opening a PR. 
 - The maintainers review the changes and merge as usual. 
 - Once maintainers have merged the features into main, they can just do this to publish a new version: 

```bash 

# Create new release
git checkout main
git pull origin main           # Get all merged changes

# Create new version tag (since we have new features)
pnpm version minor            # This creates a git tag like this v1.1.0 for example
git push --follow-tags        # This triggers GitHub Action to publish
```

Once we push the new version tag, the workflow will get triggered which will publish the new version to npm. The releases in the Github UI is basically for documentation purposes and informing users. 

This way we do not need to create a new release branch and don't need to keep the PRs open and instruct devs to bump the versions themselves thus giving us and the devs more flexibility.

I will update the publish workflow to reflect this new change.

m2rads:
For your reference here's a complete flow with developers and maintainers perspective: 

```bash

# INITIAL SETUP BY MAINTAINER (one-time setup)
git checkout main
git pull origin main
pnpm version minor      # Creates initial tag v1.0.0
git push --follow-tags  # This publishes v1.0.0

# DEVELOPER 1 WORKFLOW
git checkout main
git pull origin main
git checkout -b feature/new-test
# Make changes...
git add .
git commit -m "feat: add new test feature"
git push -u origin feature/new-test
# Create PR on GitHub

# DEVELOPER 2 WORKFLOW (working simultaneously)
git checkout main
git pull origin main
[...]

PR #210, opened 2024-12-30
update publish workflow

m2rads:
This PR addresses issue #168 

### Changes
- Updated publish workflow to trigger the release based on the tag instead of merges to the main branch

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -2,10 +2,8 @@ name: Publish Package
 
 on:
   push:
-    branches:
-      - main
-    paths:
-      - 'packages/shortest/**'
+    tags:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -2,10 +2,8 @@ name: Publish Package
 
 on:
   push:
-    branches:
-      - main
-    paths:
-      - 'packages/shortest/**'
+    tags:
+      - 'v*'
```
m2rads, line 6: This will now look for the newly created tag like `v0.2.1` and once the tag is pushed to main like `git push --follow-tag` it will trigger the workflow to publish the package to npm registery.

Issue #180, opened 2024-12-26
Automated building and execution


gladyshcodes:
### What

Create a script similar to `nodemon` that would listen to changes and rebuild the project for us on the fly.

### Why

Now, in order to test newly made changes, you will need to build `packages/shortest`, and then re-run tests. That takes some development time.

m2rads:
I think this could improve the developer experience. If you end up using a new lib for this, please be sure to add it as devDependency.

PR #216, opened 2024-12-30
Expose package watch mode

gladyshcodes:
Issue: #180 

**Proposed change**
`dev` script was updated to auto-build package. This removes the burden to manually run `pnpm build` after every change made to `/packages/shortest`

**Why**
Improve DX


slavingia (approved):

slavingia:
Some merge conflicts to address!

slavingia:
Addressed merge conflicts.

packages/shortest/package.json
```diff
@@ -29,7 +29,7 @@
     "build:js": "esbuild src/index.ts --bundle --platform=node --format=esm --outfile=dist/index.js --external:esbuild --external:punycode --external:playwright --external:@anthropic-ai/sdk --external:expect --external:dotenv",
     "build:cjs": "esbuild src/index.ts --bundle --platform=node --format=cjs --outfile=dist/index.cjs --external:esbuild --external:punycode --external:playwright --external:@anthropic-ai/sdk --external:expect --external:dotenv",
     "build:cli": "esbuild src/cli/bin.ts --bundle --platform=node --format=esm --outdir=dist/cli --metafile=dist/meta-cli.json --external:fsevents --external:chokidar --external:glob --external:esbuild --external:events --external:path --external:fs --external:util --external:stream --external:os --external:assert --external:url --external:playwright --external:@anthropic-ai/sdk --external:expect --external:dotenv --external:otplib --external:picocolors --external:punycode --external:https --external:http --external:net --external:tls --external:crypto --external:mailosaur",
-    "dev": "pnpm build:types --watch",
+    "dev": "pnpm build --watch",
```
m2rads, line 32: I think we should update the `package.json` in the root to run this script otherwise users would need to do `cd packages/shortest` every time which is very annoying. Maybe you could update the `build:pkg` to run this script instead of `pnpm build`. Also I think the naming should be something like `watch:build` instead of dev.

m2rads: On a second thought, I think it's better to have a script in root `package.json` like this 
```typescript 
"watch:build": "cd packages/shortest && pnpm dev"
``` 
I think this would be more appropriate.

gladyshcodes: Good point!

gladyshcodes:
@rmarescu fyi I am going to make last commit and merge this pr tomorrow cet

rmarescu:
@gladyshcodes have you tried using [`pnpm shortest:dev`](https://github.com/anti-work/shortest/blob/cd4a7875528f0404cf2bb8a54d46f14a0c15a558/package.json#L22) script, that can be used to test the CLI without build?

gladyshcodes:
> @gladyshcodes have you tried using [pnpm shortest:dev](https://github.com/anti-work/shortest/blob/cd4a7875528f0404cf2bb8a54d46f14a0c15a558/package.json#L22) script, that can be used to test the CLI without build?

@rmarescu I believe you're referring to slightly different thing. this pr isn't about running the CLI with `npx` without a build; rather, it's about watching build changes in package in development mode, so as not to do it manually

gladyshcodes:
pr is RFM

slavingia:
RFM? Let's avoid acronyms not everyone will know.

gladyshcodes:
> RFM? Let's avoid acronyms not everyone will know.
Ready for merge :D

package.json
```diff
@@ -6,20 +6,23 @@
     "dev": "next dev --turbo",
```
rmarescu, line 6: All app-related scripts.

package.json
```diff
@@ -6,20 +6,23 @@
...
-    "setup": "npx tsx lib/setup.ts",
     "db:seed": "npx tsx lib/db/seed.ts",
     "db:generate": "drizzle-kit generate",
     "db:migrate": "npx tsx lib/db/migrate.ts",
     "db:studio": "drizzle-kit studio",
     "db:cleanup": "tsx lib/db/cleanup.ts",
     "stripe:webhooks": "stripe listen --forward-to http://localhost:3000/api/stripe/webhook",
+    "setup": "npx tsx lib/setup.ts",
+
+    "pkg:build": "cd packages/shortest && pnpm build",
```
rmarescu, line 17: All package-related scripts, prefixed with `pkg`.

package.json
```diff
@@ -6,20 +6,23 @@
...
+    "pkg:dev": "cd packages/shortest && pnpm dev",
+    "pkg:test": "cd packages/shortest && pnpm shortest",
+    "pkg:test:src": "npx tsx packages/shortest/src/cli/bin.ts",
+
     "test": "pnpm shortest",
-    "prepare": "cd packages/shortest && pnpm build",
```
rmarescu, line 18: Dupe of `pkg:build`.

rmarescu (approved):
Added a couple of more changes.

PR #224, opened 2024-12-31
Add 'before' function

slavingia:


packages/shortest/index.d.ts
```diff
@@ -26,13 +26,14 @@ declare module "@antiwork/shortest" {
     expect(description: string): TestChain;
     expect(
       description: string,
-      fn?: (context: TestContextProps) => Promise<void>,
+      fn?: (context: TestContextProps) => Promise<void>
```
slavingia, line 29: My editor is doing this even though I have eslint/prettier autosave; any ideas? Doesn't do the same in other files. Also, shouldn't auto fix auto-correct these?

app/__tests__/dashboard.test.ts
```diff
@@ -1,6 +1,4 @@
 import { shortest } from "@antiwork/shortest";
 
-shortest(
-  "Login to the App using magic link. Use this email: 'mission-health@rdt7stzf.mailosaur.net'",
-);
+shortest("Login using this email: mission-health@rdt7stzf.mailosaur.net");
```
slavingia, line 3: More natural. The more it looks like it's written for humans the better imo

app/__tests__/dashboard.test.ts
```diff
@@ -1,6 +1,4 @@
...
 shortest("Verify that the user can access the /dashboard page");
```
slavingia, line 4: Once we have chaining, we should update the tests to chain!

packages/shortest/index.d.ts
m2rads: looks like the , character is consistently removed in this file. I think for this we need to add a new rule in eslint.

m2rads (commented):
Looking good.

packages/shortest/index.d.ts
```diff
@@ -2,7 +2,7 @@ import type { Expect } from "expect";
 import type { Page, Browser, APIRequest, APIRequestContext } from "playwright";
 import type * as playwright from "playwright";
 import type { ShortestConfig } from "./dist/types/config";
-import type { TestAPI, TestContext } from "./dist/types/test";
```
slavingia, line 5: This was triggering a linting error.

slavingia:
If anyone else wants to review, lmk!

packages/shortest/index.d.ts
slavingia: @m2rads I'm not sure about this change. Also shortest is failing on this branch. Imo the #1 priority should be stability before additional features.

slavingia: Should be fixed once this is merged: https://github.com/anti-work/shortest/pull/231

packages/shortest/index.d.ts
amk-dev: might be conflicts between eslint and prettier settings.

on the root of the project, create a `.prettierrc` and add the below setting. toggle the "all" / "none" based on if its adding / removing the comma.

probably a good idea to commit this also

```
{
  "trailingComma": "all"
}
```

@slavingia

amk-dev: ohh shit, saw its already added. nevermind

Issue #217, opened 2024-12-30
Better way to ensure DOM is settled


gladyshcodes:
### What

Get rid of `setTimeout` and use Playwright API instead

### Why

Currently, we use timeouts, wait one second before performing any actions on a page: `new Promise((resolve) => setTimeout(resolve, ms));`

This is ineffective for several reasons:
- DOM settlement can potentially take more than 1 second, making tests flaky
- Performance decrease: setTimeout unnecessarily introduces a delay, even when the DOM settles faster

slavingia:
Eventually I'd like to deprecate Playwright and rely on Computer Use; so a "native" solution would be cool. We could always start with PW and then eventually deprecate.

gladyshcodes:
Understood. I will pick it up shortly and implement it without Playwright

Asking about your vision—are you aiming for Shortest to eventually become the "next-gen Playwright"?

slavingia:
Yep!

"Playwright can automate browser tasks in Chromium, Firefox, and WebKit with a single API."

Eventually computer use will be able to do all of this, and there won't be a need for playwright. I think we should bet on this!

gladyshcodes:
> Eventually computer use will be able to do all of this, and there won't be a need for playwright. I think we should bet on this!

Agree!

PR #225, opened 2024-12-31
Ensure DOM is settled

gladyshcodes:
**Issue #217** 

**Proposed change**
Get rid of `setTimeout` brute force

**Why**
Ineffective (see why in the issue)

**Outcome / numbers:**
Process that took as a 1s to perform now takes about **1-5 ms** (that's how much it actually takes for [DOMContentLoaded](https://developer.mozilla.org/en-US/docs/Web/API/Document/DOMContentLoaded_event) event to be emitted)

However, since the pages Shortest navigates through may experience DOM mutations even after the initial page load (e.g., when modals are opened), we continue to wait until all DOM mutations are completed using `MutationObserver`


gladyshcodes:
@slavingia FYI Currently, implementing such functionality without Playwright will be challenging and time-consuming as Playwright has its own internal lifecycle events that it emits / consumes

When trying to do this natively I receive:
``` bash
Failed to wait for stable DOM: page.evaluate: Error: Timed out after 1000ms waiting for the DOM to stabilize.
at eval (eval at evaluate (:234:30), <anonymous>:5:15)
```

[See how Playwright listens to events](https://github.com/microsoft/playwright/blob/3bd97776b8473713a13d8a70be45e10c17af0dda/src/rpc/client/frame.ts#L150)
[See how Playwright emits events](https://github.com/microsoft/playwright/blob/cdb862767f408fa6d364179d449f562565b4a793/packages/playwright-core/src/client/frame.ts#L71-L82)

I think we could always start with Playwright and then deprecated it with `.evaluate` and other deps altogether

slavingia:
Sounds good, we can keep playwright for now!

m2rads:
Looks good! Can we add this method for Github and Mailosaur tool as well? @gladyshcodes

gladyshcodes:
I guess I will close this PR as many fundamental changes will be made to those files in upcoming PR for issue #227. FYI @slavingia @m2rads 

> Looks good! Can we add this method for Github and Mailosaur tool as well? @gladyshcodes

Okay, will make sure it's used everywhere across the codebase

PR #228, opened 2024-12-31
feat: enable bash tool for computer use

### LINKED ISSUE #226: cURL support via bash tool usage
### Why

[Anthropic's computer use supports a bash tool](https://docs.anthropic.com/en/docs/build-with-claude/computer-use#bash-tool). We should be using this so that we can have the client run cURL commands, in order to test API functionality in an end-to-end manner.

Bounty: $2,000 (you'll be able to take this in cash, Anti-Work (Gumroad, Inc.) options, or a combo).

amk-dev:
Fixes #226 

**Changes**

1. This PR adds support for the `bash` tool that's supported by anthropic computer use.
2. For cached tests having bash tools, they're rerun without caching

https://github.com/user-attachments/assets/ac477074-976f-4dfe-ab60-330967ec2096

### DIFF
--- a/examples/curl.test.ts
+++ b/examples/curl.test.ts
@@ -1,0 +1,5 @@
+import { shortest } from "@antiwork/shortest";
+
+shortest("run a request to /posts/1 and verify the post is created by user 1", {
+  url: "https://jsonplaceholder.typicode.com",
+});
--- a/packages/shortest/src/ai/client.ts
+++ b/packages/shortest/src/ai/client.ts
@@ -130,7 +130,10 @@ export class AIClient {
             ) {
               return {
                 toolBlock,
-                result: runBashCommand(input.command),
+                result: browserTool.execute({
+                  action: "bash",
+                  command: input.command,
+                }),
               };
             }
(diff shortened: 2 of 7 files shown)


amk-dev:
right now if a command is not found, in this case `jq`, ai automatically tries to install it, should this be allowed ?

@slavingia 

image

slavingia:
I think we should probably constrain it a few commands that will get regular usage and add procedurally as there are needs.

amk-dev:
thinking about caching,

any step from the bash command might need to be cache invalidated. because these values can change the results of  the upcoming steps.

caching the bash result also doesnt make sense, because they're not static.

slavingia:
Option 2 makes sense. We can make it more sophisticated over time. For example we probably shouldn't cache the CI much, but locally may want to.

slavingia:
Merged in one bash tool PR, if you think there are improvements still to be had, please merge main onto this branch and just make those changes. Will distribute bounty fairly as soon as we have the Flexile feature built to do so!

PR #231, opened 2025-01-01
chore: improve index.d.ts generation

### LINKED ISSUE #205: Generating index.d.ts automatically when bundling types
### The issue: 

`src/index.ts` is our source file that contains the actual implementation and gets compiled to JavaScript.

`index.d.ts` is our declaration file that only contains type definitions and Is used by TypeScript for type checking. It also defines the public API types for consumers.

We need both because the compiled JavaScript `(dist/index.js)` doesn't contain type information and external users of our package need type definitions for TypeScript support.

### Why we need to fix it: 
Having to change both files manually makes the contributions to this project more error prone. 

### Here's how we can fix it: 
This update to `package.json` should most likely fix this issue: 

```typescript 
{
  "types": "./dist/types/index.d.ts",  // Point to generated types
  "build:types": "tsc --declaration --emitDeclarationOnly --outDir dist/types"  // Generate declarations
}
```

amk-dev:
Fixes #205, #191

**Changes**

1. Update the `index.d.ts` location from `index.d.ts` to `dist/index.d.ts`
2. Generate `index.d.ts` automatically  
3. Remove the manually written `index.d.ts`
3. Add `tsup` as a dev dep, 
    * this tool is used to generate the `index.d.ts` 
    * this can also be used to bundle the library, its build with `esbuild` under the hood, and has support for `@microsoft/api-extractor` which i wanted to use previously.
    * not using any `api-extractor` features, but it can be used if the need arises.
4. Augment global to have `expect` included in the declaration files
5. Make `anthropicKey` and `mailosaur` fields & child fields optional in `ShortestConfig`, because they can come from env variables

Note: in some previous commits, i can see `tsup` was used initially for bundling and then removed. that's why i kept it only for generating types. any reason for this ?


packages/shortest/src/types/config.ts
```diff
@@ -2,7 +2,7 @@ export interface ShortestConfig {
   headless: boolean;
   baseUrl: string;
   testDir: string | string[];
-  anthropicKey: string;
+  anthropicKey?: string;
```
slavingia, line 5: This should be required IMO.

amk-dev: I've had the same opinion. but when i checked `OpenAI` / `Anthropic` sdks, the apiKey was optional. kinda makes sense because they allow loading them from the environment. and when i looked shortest's code has a fallback to the env variable. 

image

but mailosaur doesnt have this fallback implemented, that's why i kept it as required.

amk-dev: We can just make it strict, and let the user handle the type errors if they need to.

for this codebase, in `shortest.config.ts`, we have multiple options,
1. augment the process.env to make these variables defined

image

2. assert that the variable is non null with (!) operator

image

slavingia: Makes sense if it can infer from the process env to have it be optional. The less required config the better imo

amk-dev: so, should i make `anthropicKey`, `mailosaur.apiKey` and `mailosaur.serverId` as optional ?

* user's can load it from `ANTHROPIC_API_KEY`, `MAILOSAUR_API_KEY`, `MAILOSAUR_SERVER_ID` variables respectively
* right now the fallback is there for anthropic key, will add them to mailosaur tool if needed

slavingia: Yes they should be optional, i think mailosaur already is

amk-dev: got it. the `mailosaur` property is optional, but their child properties are not. will update them.

image

slavingia: Ah that's confusing / a bug imo.

amk-dev: done ✅

packages/shortest/src/types/config.ts
```diff
@@ -1,10 +1,10 @@
 export interface ShortestConfig {
   headless: boolean;
   baseUrl: string;
-  testPattern: string;
-  anthropicKey: string;
+  testDir: string | string[];
```
slavingia, line 4:
```suggestion
  testPattern: string;
```

slavingia (changes requested):
Getting an autofix linting error.

amk-dev:
type error resolved.

is #238 taking precedence over this ? @m2rads 
this PR takes care of the global export also, to keep the current

PR #233, opened 2025-01-01
Bash tool

gladyshcodes:
**Issue #226**

This PR introduces a Bash tool that enables using Shortest for API testing


gladyshcodes:
A couple of things:

gladyshcodes:
Page extraction. To extract data such as bearers, I propose three solutions

1. Add extract method, providing it with schema of data we want to extract
``` js
const apiKey = await shortest.extract({
  instruction: "extract API key“,
  schema: z.object({
    username: z.string(),
    url: z.string(),
  }),
});
```

2. Leverage shortest directly with an ExtractSchema object, keeps things simple:
``` js
const apiKey = await shortest(“extract API key”, new ExtractSchema{
  schema: z.object({
    username: z.string(),
    url: z.string(),
  }),
});
```

3. Third, use FormData to create structured requests
``` js
const keys = new FormData(); // uses shortest behind the hood
keys.set('bearer', ‘extract bearer token);
```

This way we are embedding values directly in the test file memory rather than relying on conversational context

Why:
-Reduced dependence on AI Memory
-Streamlined code, allows us to access vars in a native way (instead of using <>)
-Potentially easier to debug
-More scalable (e.g what if we need to save those values somewhere in the future?)

gladyshcodes:
@slavingia Test cases with API requests that we would like to fully migrate to Shortest, in my opinion, have one drawback: API queries like that are super inconsistent and unstructured

In this PR, I propose an `ApiRequest` class (inspired by Playwright's API design). This class standardizes the format of requests to LLMs by aligning them with the [RequestInit](https://developer.mozilla.org/en-US/docs/Web/API/RequestInit) structure from fetchAPI

Why:
-API request queries are inconsistent and unstructured
-Inconsistency potentially leads to decrease in precision. Since RequestInit is the most widely used format, there's a good chance LLM will be more precise when working with it
-Lack of structure makes tool less readable and harder to use

gladyshcodes:
Had a chance to refactor `packages/shortest/src/ai/client.ts`, now it's far more readable

gladyshcodes:
I am unable to test the functionality in the final test file of #226. I'm encountering this issue when trying to create an organization at Iffy:
Email: sort-better@wxxzt0jh.mailosaur.net

Screenshot 2025-01-01 at 15 05 14

@slavingia Could you assist me with that?

slavingia:
It may be good to add an endpoint within shortest and test for that so we can have it as part of this repo's tests/CI?

gladyshcodes:
> It may be good to add an endpoint within shortest and test for that so we can have it as part of this repo's tests/CI?
Done

packages/shortest/src/ai/client.ts
```diff
@@ -113,95 +120,120 @@ export class AIClient {
...
-
-          const toolResults = toolBlocks.map((toolBlock) => {
-            return {
-              toolBlock,
-
-              result: browserTool.execute(toolBlock.input as any),
-            };
-          });
-
-          const results = await Promise.all(toolResults.map((t) => t.result));
```
gladyshcodes, line 129: `Promise.all` is not required here and therefore removed. We will only have one executable tool per response if its stop_reason is tool_use

amk-dev: Hey i don't this is true. here's an example response with multiple tools,

```
{
    "id": "msg_019cgXeAnYKA7GSMVciH3scQ",
    "type": "message",
    "role": "assistant",
    "model": "claude-3-5-sonnet-20241022",
    "content": [
        {
            "type": "text",
            "text": "I'll help you check both the weather in New York and the current time there."
        },
        {
            "type": "tool_use",
            "id": "toolu_019nyJLidjoEMwsGPNXuCVkW",
            "name": "get_weather",
            "input": {
                "location": "New York, NY"
            }
        },
        {
            "type": "tool_use",
            "id": "toolu_019QJVhaHd9HQCs2ZFmK1N4Y",
            "name": "get_time",
            "input": {
                "timezone": "America/New_York"
            }
        }
    ],
    "stop_reason": "tool_use",
    "stop_sequence": null,
    "usage": {
        "input_tokens": 526,
        "cache_creation_input_tokens": 0,
        "cache_read_input_tokens": 0,
        "output_tokens": 114
    }
}
```

here's the curl for this

```sh
curl https://api.anthropic.com/v1/messages \
     --header "x-api-key: $ANTHROPIC_API_KEY" \
     --header "anthropic-version: 2023-06-01" \
     --header "content-type: application/json" \
     --data \
'{
    "model": "claude-3-5-sonnet-20241022",
    "max_tokens": 1024,
    "tools": [{
        "name": "get_weather",
        "description": "Get the current weather in a given location",
        "input_schema": {
            "type": "object",
            "properties": {
                "location": {
                    "type": "string",
                    "description": "The city and state, e.g. San Francisco, CA"
                },
                "unit": {
                    "type": "string",
                    "enum": ["celsius", "fahrenheit"],
                    "description": "The unit of temperature, either 'celsius' or 'fahrenheit'"
                }
            },
            "required": ["location"]
        }
    },
    {
        "name": "get_time",
        "description": "Get the current time in a given time zone",
        "input_schema": {
            "type": "object",
            "properties": {
                "timezone": {
                    "type": "string",
                    "description": "The IANA time zone name, e.g. America/Los_Angeles"
                }
            },
            "required": ["timezone"]
        }
    }],
    "messages": [{
        "role": "user",
        "content": "What is the weather like right now in New York? Also what time is it there?"
    }]
}'
```

gladyshcodes: My bad! I have fixed that, thank you for seeing it.

This also helped me to spot one sneaky bug.

gladyshcodes:
So this thing is R4R

Please take a look @slavingia @m2rads

examples/api-assert-bearer.test.ts
```diff
@@ -0,0 +1,16 @@
...
+  Test the API POST endpoint ${TESTING_API_BASE_URI}/assert-bearer with body { "flagged": "false" } without providing a bearer token.
+  Expect the response to indicate that the token is missing
+`;
+
+const validBearerToken = `
+  Test the API POST endpoint ${TESTING_API_BASE_URI}/assert-bearer with body { "flagged": "true" } and the bearer token ${ALLOWED_TEST_BEARER}.
+  Expect the response to show "flagged": true
+`;
+
+shortest(missingBearerToken);
```
slavingia, line 15: Not sure the variables add much, can just put the strings in here?

slavingia (commented):
Minor comments. We should update the README.

examples/api-assert-bearer.test.ts
gladyshcodes: Sure

gladyshcodes:
@slavingia README updated

slavingia:
Would be good to have an example within this repo itself, somewhere in here: https://github.com/anti-work/shortest/tree/main/app/api

Of a test. I'd also think we'd want to take advantage of the "after" async page block thingy so we can verify the output more specifically, than just "valid users" – how does the QA AI/person know that the specific users were listed appropriately, and none that should be hidden, aren't shown (e.g. the app allows one to have a deleted user, that may still show up)?

gladyshcodes:
> Would be good to have an example within this repo itself, somewhere in here: https://github.com/anti-work/shortest/tree/main/app/api

Not sure what you mean by 'putting examples into the repo itself'. We now have three more working tests in `/examples` folder:

[examples/api-failure.test.ts](https://github.com/anti-work/shortest/pull/233/files#diff-b12db0b03698267ca9c7895dec0a03a0f90361db0e14d231c39dbe37357a9bd0)
[examples/api-assert-bearer.test.ts](https://github.com/anti-work/shortest/pull/233/files#diff-438245a1c8df3205316ed68c9215efa6520fcfe4dcb0a362e4e16b7d69137fba)
[examples/api-assert-bearer-2.test.ts](https://github.com/anti-work/shortest/pull/233/files#diff-cc1db6a58a2e9d729c6b54afe97646194453d9367f2daf5c1c64abc4cd1b78c9)

Have you seen them?
You mean you want those tests to be near to test API endpoint file themselves (route.test.ts)?

README example is made up, we do not have such test in our codebase

slavingia:
My bad, I missed those.

m2rads:
Looks solid. Let's have these tests in the `api/test/assert-bearer` token so that we can have them run in the CI.

PR #238, opened 2025-01-02
Auto generate index.d.ts by typescript

m2rads:
This PR addresses issue #231 

### Changes

- Update `build:types` script to generate types using tsc
- Refactor `index.ts` to import `expect`
- Update test files to reflect the new changes
- Update readme file


README.md
```diff
@@ -55,7 +55,7 @@ yarn shortest
     headless: false,
     baseUrl: 'http://localhost:3000',
     testPattern: '**/*.test.ts',
-    anthropicKey: process.env.ANTHROPIC_API_KEY
+    anthropicKey: process.env.ANTHROPIC_API_KEY || '',
```
m2rads, line 58: This is now necessary as typescript is considering env vars as `string | undefined` therefore we have to add an fallback empty string to silence typescript errors. As mentioned before we can have these managed internally without having users to declare these here.

README.md
```diff
@@ -72,7 +72,7 @@ You can also use callback functions to add additional assertions and other logic
 execution in browser is completed.
 
 ```typescript
-import { shortest } from '@antiwork/shortest';
+import { shortest, expect } from '@antiwork/shortest';
```
m2rads, line 75: One of the main reasons why we created `index.d.ts` manually was to allow users to use jestExpect without having to import it. With the new changes this is necessary.

packages/shortest/package.json
```diff
@@ -5,28 +5,27 @@
   "type": "module",
   "main": "./dist/index.js",
   "module": "./dist/index.js",
-  "types": "./index.d.ts",
+  "types": "./dist/types/index.d.ts",
```
m2rads, line 8: reading the autogenerated types instead of the file we created manually

packages/shortest/package.json
```diff
@@ -5,28 +5,27 @@
...
-    "index.d.ts"
+    "dist/cli"
   ],
   "scripts": {
     "build": "rm -rf dist && pnpm build:types && pnpm build:js && pnpm build:cli",
     "prepare": "pnpm build",
     "prepublishOnly": "pnpm build",
     "postinstall": "node -e \"if (process.platform !== 'win32') { try { require('child_process').execSync('chmod +x dist/cli/bin.js') } catch (_) {} }\"",
-    "build:types": "tsc --emitDeclarationOnly --outDir dist/types && cp index.d.ts dist/",
+    "build:types": "tsc --emitDeclarationOnly --outDir dist/types",
```
m2rads, line 28: This script is in charge of generating the types.

packages/shortest/src/index.ts
```diff
@@ -227,6 +224,7 @@ export const test: TestAPI = Object.assign(
   },
 );
 
+export const expect = jestExpect;
```
m2rads, line 227: Making jestExpect available as an import in the test files

packages/shortest/src/types/globals.ts
```diff
@@ -5,9 +5,21 @@ export interface ShortestGlobals {
...
+
+declare global {
+  namespace NodeJS {
+    interface Global {
+      __shortest__: ShortestGlobals;
+    }
+  }
+}
+
+export {};
```
m2rads, line 25: Part of jest setup. we had it in `index.d.ts` before but now we declare it here. This will get autogenerated by typescript.

shortest.config.ts
```diff
@@ -4,9 +4,9 @@ export default {
   headless: false,
   baseUrl: "http://localhost:3000",
   testPattern: "**/*.test.ts",
-  anthropicKey: process.env.ANTHROPIC_API_KEY,
+  anthropicKey: process.env.ANTHROPIC_API_KEY || "",
   mailosaur: {
-    apiKey: process.env.MAILOSAUR_API_KEY,
-    serverId: process.env.MAILOSAUR_SERVER_ID,
+    apiKey: process.env.MAILOSAUR_API_KEY || "",
+    serverId: process.env.MAILOSAUR_SERVER_ID || "",
```
m2rads, line 10: Have to do this otherwise typescript throws an error.

m2rads (commented):
Added self-review. I can further cleanup the types to improve DX further.

slavingia:
If the expect changes are necessary, would be good to do in its own PR. cc @amk-dev

slavingia:
Would be good to resolve merge conflicts now that the other branch was merged. I personally don't think doing || "" is a great idea, it pollutes the README/config and looks quite ugly. I'm sure there's a more elegant solution. Importing `expect` isn't a big deal, especially since we don't really need expect since the shortest itself should know what to look for to pass. So it'll act like a smell if it's included at all.

m2rads:
I think for the env vars we can just have the user set them up in their .env file and we directly read them from there. I think Vercel AI SDK is like that. 

For expect I had to update it in this PR otherwise it would have introduced a bug to the package. Since we already have expect implemented, if a user upgrade their shortest, their implementation would be broken.

If you don't think expect is necessary, I can delete the jestExpect implementation all together.

slavingia:
Makes sense to me to remove it.

amk-dev:
> I personally don't think doing || "" is a great idea, it pollutes the README/config and looks quite ugly. I'm sure there's a more elegant solution. 

the merged PR should take care of this.

> Importing expect isn't a big deal, especially since we don't really need expect since the shortest itself should know what to look for to pass. So it'll act like a smell if it's included at all.

the merged PR preserves the current behaviour. `expect` is globally available. didn't want to introduce a breaking change.

---

one major difference from this PR to the merged PR is the use of `tsup`. its a pretty standared tool used on many popular projects. has a lot of features, but we're not using any atm apart from that global expect. but if you guys want to remove it, that's also fine.

amk-dev:
> I think for the env vars we can just have the user set them up in their .env file and we directly read them from there. I think Vercel AI SDK is like that.

had a discussion in the last PR about this.

https://github.com/anti-work/shortest/pull/231#discussion_r1900478123

after my PR merge, 
-> any field that can accept an env variable is optional. the code has a fallback to the env variable.

m2rads:
> > I think for the env vars we can just have the user set them up in their .env file and we directly read them from there. I think Vercel AI SDK is like that.
> 
> had a discussion in the last PR about this.
> 
> [#231 (comment)](https://github.com/anti-work/shortest/pull/231#discussion_r1900478123)
> 
> after my PR merge, -> any field that can accept an env variable is optional. the code has a fallback to the env variable.

Just saw your PR. This PR is not required anymore as your solves it in a better way.

Issue #244, opened 2025-01-02
Release new version: 0.4.0


slavingia:
### What

- Bash tool
- Chaining tests
- Bug fixes
- Test pattern versus test dir

slavingia:
Made it both minor and patch as it has new features and breaking changes.

m2rads:
Is the bug regarding chaining resolved now?

slavingia:
I think that was just due to not pushing the release, so the README is out of date

m2rads:
I think the next version should be v0.4.0. 

So if I bump the version with `pnpm version mintor` it would create a tag for v0.4.0. For v0.4.1 we have to do another `pnpm version patch` after that so we get v0.4.1. This means that in order to go from v0.3.0 to v0.4.1 we have two releases.

slavingia:
updated title to 0.4.0

m2rads:
Thanks. I will release v.0.4.0 now

m2rads:
Published: 

![image](https://github.com/user-attachments/assets/8e9d4e6b-719c-4b59-9f13-8d83a5f0578b)

PR #245, opened 2025-01-02
Update npm readme

m2rads:
Issue #244 

In prepration for release of v0.4.0, we need to update the README that will be shown in npm website.


packages/shortest/README.md
```diff
@@ -8,61 +8,76 @@ AI-powered natural language end-to-end testing framework.
...
-## Quick Start
+Run tests:
+
+```bash
+npx shortest
+# or
+pnpm shortest
+```
+
+### Quick start
```
slavingia, line 33: Think we should remove this, as it's mentioned below and technically can't do this until the rest is done?

m2rads: Done

PR #252, opened 2025-01-03
feat: add baseURL configuration to BrowserManager context

PedroAVJ:
I figured it makes sense that, page.goto, should allow for relative paths like "/dashboard" instead of having to define the full url. It assumes that relative paths are using the baseUrl defined in the config.


m2rads:
Makes sense

m2rads (approved):

PR #254, opened 2025-01-04
Improve AI usage of newly open tabs

rmarescu:
* Automatically focus on newly open tabs
* Remove hard-coded Mailosaur email address 
  * The email is AW-specific and cannot be used by external contributors
  * Remove the need of having another ENV for the email by using a standard format for the email: `shorters@<MAILOSAUR_SERVER_ID>.mailosaur.net`
* Improve README instructions for Mailosaur
* Remove flaky test with a more deterministic one


app/(dashboard)/page.test.ts
```diff
@@ -1,5 +1,8 @@
 import { shortest } from "@antiwork/shortest";
 
+const loginEmail = `shortest@${process.env.MAILOSAUR_SERVER_ID}.mailosaur.net`;
+if (!loginEmail) throw new Error("MAILOSAUR_LOGIN_EMAIL is required");
+
 shortest("Verify that buttons on the landing page are rounded");
-shortest("Login using this email: mission-health@rdt7stzf.mailosaur.net");
+shortest("Log in", { email: loginEmail });
 shortest("Verify that the user can access the /dashboard page");
```
rmarescu, line 8: This test doesn't seem to work as the user with Mailosaur email doesn't have GitHub connected, and the AI is confused by the Dashboard page:

CleanShot 2025-01-03 at 20 11 52@2x

Unsure what would be a good way to fix it. Maybe change the test, or just delete for now?

app/(dashboard)/page.test.ts
```diff
@@ -1,5 +1,8 @@
 import { shortest } from "@antiwork/shortest";
 
+const loginEmail = `shortest@${process.env.MAILOSAUR_SERVER_ID}.mailosaur.net`;
+if (!loginEmail) throw new Error("MAILOSAUR_LOGIN_EMAIL is required");
+
```
m2rads, line 5: Why not add the whole email as env var? And then we can add that to Github secrets too so that the workflow can run. 

Also, Would be a good idea to add server_id, api_key and email address to the Vercel env var so that Antiwork team members can set up their local more easily. 

```typescript     
 MAILOSAUR_EMAIL: ${{ secrets.MAILOSAUR_EMAIL }}
```

m2rads (approved):

app/(dashboard)/page.test.ts
slavingia: We can delete the test for now.

rmarescu:
The [test is failing](https://github.com/anti-work/shortest/actions/runs/12610803403/job/35145836872?pr=254) as AI seems to be confused on using the new tab after clicking the `Sign in` button on the email tab.

During the run below, `screenshot-2025-01-04T14-12-54-219Z.png` is actually the email tab, and then is trying to use keyboard shortcuts to navigate tabs, which are Windows-specific.

Attempting to adjust the system prompt to handle this better.

CleanShot 2025-01-04 at 06 13 26@2x

slavingia:
Weird as it was working before with the old prompt I believe

rmarescu:
I've executed the same test from `main`, and the behaviour is quite different. It passed 2/4 times:

<details>
<summary>Run 1: passed once it checked the email was received and had a link (did not click on it)</summary>

CleanShot 2025-01-04 at 06 44 24@2x
</details>

<details>
<summary>Runs 2&3: failed because couldn't switch to the new tab after clicking the email link (kept retrying by opening new tabs 5 times)</summary>

All the screenshots taken were of the email tab.

CleanShot 2025-01-04 at 06 47 51@2x

</details>

<details>
<summary>Run 4: passed after clicking the email link, didn't check the new page loaded</summary>

CleanShot 2025-01-04 at 06 53 57@2x


</details>

packages/shortest/src/browser/core/browser-tool.ts
```diff
@@ -62,6 +62,12 @@ export class BrowserTool extends BaseBrowserTool {
     this.viewport = { width: config.width, height: config.height };
     this.testContext = config.testContext;
 
+    // Update active page reference to a newly opened tab
+    this.page.context().on("page", async (newPage) => {
```
rmarescu, line 66: This fixes the issue reported, where the AI didn't have the new tab, after clicking on the email link, as active.

@m2rads do you think this could cause any possible side effects?

PR #256, opened 2025-01-04
Make Shortest compatible with Windows 

### LINKED ISSUE #251: Unable to run because of failure to create output directory
I'm trying to run shortest test from my folder in D:, however if fails to start with the following error. Where do I set a custom output directory? or should I be running it from C: only?

```
D:\work\shortestai\app\__tests__\login.test.ts
X [ERROR] Failed to create output directory: mkdir C:\Users\amit\AppData\Local\Temp\shortest-cache\D:: The filename, directory name, or volume label syntax is incorrect.

X [ERROR] Failed to create output directory: mkdir C:\Users\amit\AppData\Local\Temp\shortest-cache\D:: The filename, directory name, or volume label syntax is incorrect.


Test Execution Error: Build failed with 2 errors:
error: Failed to create output directory: mkdir C:\Users\amit\AppData\Local\Temp\shortest-cache\D:: The filename, directory name, or volume label syntax is incorrect.
error: Failed to create output directory: mkdir C:\Users\amit\AppData\Local\Temp\shortest-cache\D:: The filename, directory name, or volume label syntax is incorrect.
⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯

 Tests    0 passed (0)
 Duration   0.29s
 Start at   7:41:15 pm

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
```

PedroAVJ:
Fixes #251 

I also had to fix some issues with page.evaluate. Apparently [__name](https://github.com/privatenumber/tsx/issues/113) is not defined due to how esbuild works, and was causing the tests to crash.

I have tested that the cli works when running pnpm build:pkg and than pnpm shortest --headless, on both windows and linux


packages/shortest/src/browser/core/browser-tool.ts
rmarescu, line 1: @PedroAVJ are the changes from this file needed for Windows compatibility?

PedroAVJ: Mmm, I don't think I did, but can't fully recall either. Either way you are right. No need to block this issue any further

packages/shortest/package.json
```diff
@@ -21,7 +21,7 @@
     "dist/cli"
   ],
   "scripts": {
-    "build": "rm -rf dist && pnpm build:types && pnpm build:js && pnpm build:cli",
+    "build": "rimraf dist && pnpm build:types && pnpm build:js && pnpm build:cli",
```
rmarescu, line 24: `rm -rf` for Node in a cross-platform implementation

rmarescu (approved):
Thanks for the contribution @PedroAVJ.
I've reverted the changes for `browser-tool.js` so we can address the issue separately.

PR #263, opened 2025-01-07
Updated reame with CLI example

Aldian1:
What:

The readme has no CLI example.

Why:

Its hard for users to understand how to setup shortest with CLI, especially if they have no technical experience doing this prior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,18 +238,13 @@ jobs:
       - name: Start dev server
         run: pnpm dev & sleep 1
         env:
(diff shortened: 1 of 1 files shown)


slavingia:
I think it would be better to link to the one in this repo? Than include it here itself.

Aldian1:
the shortest.yaml? is that all you need inside the config to run?

slavingia:
.github/workflows/shortest.yml

Aldian1:
i am so dumb lol. added a link instead to exisiting example.

PR #268, opened 2025-01-11
feat(cli): Use Anthropic via Amazon Bedrock

taylor-hudson:
# Add AWS Bedrock Support

Adds support for using AWS Bedrock as an alternative to the direct Anthropic API.

### Direct Anthropic API (Default)
shortest.config.ts
```typescript
export default {
  anthropicKey: process.env.ANTHROPIC_API_KEY 
} satisfies ShortestConfig;
```

### AWS Bedrock
```typescript
export default {
  useBedrock: true,
  awsAccessKey: "your-aws-access-key", // or AWS_ACCESS_KEY_ID
  awsSecretKey: "your-aws-secret-key", // or AWS_SECRET_ACCESS_KEY  
  awsRegion: "us-east-1" // or AWS_REGION
}
```

Existing implementations using direct Anthropic API will continue to work without changes.


packages/shortest/src/ai/client.ts
```diff
@@ -9,22 +10,36 @@ import { SYSTEM_PROMPT } from "./prompts";
 import { AITools } from "./tools";
 
 export class AIClient {
-  private client: Anthropic;
+  private client: Anthropic | AnthropicBedrock;
```
gladyshcodes, line 13: Thank you for contributing, Taylor!

One thing here: do `Anthropic` and `AnthropicBedrock` share different interfaces? If so, it will be nice to create abstract class around them to make our code more scalable when we decide add more AI clients?

taylor-hudson: this is a good idea. i wrote this commit originally for my personal fork but saw someone else wanted it as well.

if i have the time, ill implement this. thanks for the feedback!

gladyshcodes: nice

rmarescu:
@taylor-hudson could you please sign the CLA so we can get this into review?

taylor-hudson:
Signed @rmarescu

packages/shortest/src/ai/client.ts
rmarescu: I've started looking into this. As new models with computer use are [around the corner](https://github.com/anti-work/shortest/discussions/288), I think it makes sense to update (at least) the config in preparation for that.

rmarescu (commented):
Does this work for you?

There are other instances where the standard `Anthropic` client is used, but `AnthropicBedrock` has a different API structure.

https://github.com/taylor-hudson/shortest/blob/3f21d71177cf848c610b2b017e46c6466e54b79a/packages/shortest/src/ai/client.ts#L51

https://github.com/taylor-hudson/shortest/blob/3f21d71177cf848c610b2b017e46c6466e54b79a/packages/shortest/src/ai/client.ts#L89

https://github.com/taylor-hudson/shortest/blob/3f21d71177cf848c610b2b017e46c6466e54b79a/packages/shortest/src/ai/client.ts#L137

etc

<hr>

We want to support other providers in the near future, and I think the right approach is to use Vercel AI SDK instead. Once that's in place, should be easier to add support for Amazon Bedrock.
For that reason, I don't think it's worth the complexity introduced by this PR at this moment.

There is an [open issue](https://github.com/anti-work/shortest/issues/291) that's ready for development, if anybody would like to give it a try.

rmarescu:
Closing for now, will revisit support for Amazon Bedrock one [Vercel AI SDK](https://github.com/anti-work/shortest/issues/291) is implemented.

rmarescu:
Added [Amazon Bedrock support](https://github.com/anti-work/shortest/issues/310) to the roadmap!

Issue #296, opened 2025-01-26
401 API error should end the test on the first try


rmarescu:
### Version

0.4.2

### Your repository URL

https://github.com/rmarescu/shortest-alert/tree/b8617d9264bb3b462f5b5f998641482b841268bc

### Actual behavior

The test is retried 3 times.

### Expected behavior

The test is terminated right away.

### System Info

```shell
System:
    OS: macOS 15.2
    CPU: (16) arm64 Apple M3 Max
    Memory: 917.05 MB / 128.00 GB
    Shell: 3.6.4 - /opt/homebrew/bin/fish
  Binaries:
    Node: 21.6.1 - ~/.asdf/installs/nodejs/21.6.1/bin/node
    npm: 10.2.4 - ~/.asdf/plugins/nodejs/shims/npm
    pnpm: 9.1.1 - ~/.asdf/shims/pnpm
  Browsers:
    Brave Browser: 132.1.74.48
    Chrome: 132.0.6834.110
    Safari: 18.2
  npmPackages:
    @antiwork/shortest: ^0.4.2 => 0.4.2
```

### Validations

- [x] I've checked for [existing issues](https://github.com/anti-work/shortest/issues)
- [x] This is a concrete bug (for Q&A use [discussions](https://github.com/anti-work/shortest/discussions))

PR #301, opened 2025-01-29
fix(cli): prevent AI API retries for select status codes

khalatevarun:
fix: #296

### DIFF
--- a/packages/shortest/src/ai/client.ts
+++ b/packages/shortest/src/ai/client.ts
@@ -53,6 +53,11 @@ export class AIClient {
           toolOutputCallback,
         );
       } catch (error: any) {
+
+        if (error.status === 401) {
+          throw error;
+        }
+        
         attempts++;
         if (attempts === maxRetries) throw error;
 
--- a/packages/shortest/src/ai/client.ts
+++ b/packages/shortest/src/ai/client.ts
@@ -43,7 +43,6 @@ export class AIClient {
   }> {
     const maxRetries = 3;
     let attempts = 0;
-    const nonRetryableErrors = [500, 401, 403];
 
     while (attempts < maxRetries) {
       try {
@@ -54,7 +53,7 @@ export class AIClient {
           toolOutputCallback,
         );
       } catch (error: any) {
-        if (nonRetryableErrors.includes(error.status)) {
+        if ([401, 403, 500].includes(error.status)) {
           throw error;
         }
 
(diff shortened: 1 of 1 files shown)


packages/shortest/src/ai/client.ts
```diff
@@ -53,6 +53,10 @@ export class AIClient {
           toolOutputCallback,
         );
       } catch (error: any) {
+        if (error.status === 401) {
```
rmarescu, line 56: I think it could have a similar behaviour for any non-succesful statuses.

rmarescu (commented):
@khalatevarun thanks for the contribution! Could you please test it against the repo linked in the original issue, and provide the output / screenshot on how the new behavior looks like?

packages/shortest/src/ai/client.ts
khalatevarun: Sure!
image


Additionally,
Just to be sure, should we throw error immediately for specific code errors like 401, 403 etc while allow retry for other code errors like 500, 502 etc?

rmarescu: 502 can let it retry, 500, 401, 403 should throw right away, less likely to self-heal on the next request.

packages/shortest/src/ai/client.ts
```diff
@@ -43,6 +43,7 @@ export class AIClient {
   }> {
     const maxRetries = 3;
     let attempts = 0;
+    const nonRetryableErrors = [500, 401, 403];
```
khalatevarun, line 46: updated logic to maintain an array of error status codes to avoid retrying @rmarescu

rmarescu: I think the array can be used directly on line 57, its purpose is self-documenting.

rmarescu, line 46:
```suggestion
```

packages/shortest/src/ai/client.ts
```diff
@@ -53,6 +54,10 @@ export class AIClient {
           toolOutputCallback,
         );
       } catch (error: any) {
+        if (nonRetryableErrors.includes(error.status)) {
```
rmarescu, line 57:
```suggestion
        if ([401, 403, 500].includes(error.status)) {
```

rmarescu (approved):

khalatevarun:
thank you!

PR #302, opened 2025-01-31
chore(cli): Improve config parsing and validation

rmarescu:
* Introduce Zod
* Config test coverage with vitest
* Improved error handling

### DIFF
--- a/packages/shortest/src/index.ts
+++ b/packages/shortest/src/index.ts
@@ -12,6 +12,7 @@ import {
   ShortestConfig,
   validateConfig,
 } from "./types";
+import { ConfigError } from "./utils/errors";
 
 // to include the global expect in the generated d.ts file
 import "./globals";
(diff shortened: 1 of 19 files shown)


packages/shortest/src/index.ts
```diff
@@ -46,70 +47,65 @@ if (!global.__shortest__) {
...
-      "  - anthropicKey: string",
-  );
+  if (configs.length === 0) {
+    throw new ConfigError(
+      "no-config",
+      "No config file found. Please create one.",
+    );
+  }
+
+  if (configs.length > 1) {
```
rmarescu, line 91: Don't allow multiple config files to co-exist (to avoid confusion).

packages/shortest/tests/test-config.ts
rmarescu, line 1: I'll create a GH workflow to run the Vitest suite in a follow-up PR. Tests pass on local.

CleanShot 2025-01-30 at 23 33 53@2x

PR #304, opened 2025-02-01
chore(cli): Add unit tests workflow

rmarescu:
* New GH workflow to run unit tests
* New test runner for existing e2e tests for the Shortest source
* New scripts within `packages/shortest`
  * `pnpm test:unit` (used by the new workflow)
  * `pnpm test:e2e` (not used at this moment)


.github/workflows/cli-test.yml
```diff
@@ -0,0 +1,34 @@
+name: CLI / Tests
+
+on: push
```
rmarescu: Runs on all commits, even directly on `main`.

.github/workflows/cli-test.yml
```diff
@@ -0,0 +1,42 @@
+name: CLI / Tests
+
+on:
+  push:
+    paths:
+      - 'packages/shortest/**'
+      - '.github/workflows/cli-*.yml'
```
rmarescu, line 7: Grouped CLI-related jobs by `cli-*` prefix to self-document their context, and to trigger checks conditionally (e.g. trigger CLI-related checks when CLI code changes)

package.json
```diff
@@ -18,11 +18,12 @@
     "cli:build": "cd packages/shortest && pnpm build",
     "cli:dev": "cd packages/shortest && pnpm dev",
     "cli:test": "cd packages/shortest && pnpm shortest",
+    "cli:typecheck": "tsc -p packages/shortest/tsconfig.json --noEmit",
 
-    "test": "pnpm shortest",
+    "nextjs:typecheck": "tsc --noEmit --incremental",
```
rmarescu, line 23: Using `cli:*` and `nextjs:*` prefixes to group by app/package.

packages/shortest/tests/e2e/index.ts
```diff
@@ -0,0 +1,96 @@
...
+  }
+  throw new Error(
+    "Could not find monorepo root (no workspace package.json found)",
+  );
+}
+
+const projectRoot = findMonorepoRoot(__dirname);
+
+/**
+ * E2E Test Runner
```
rmarescu, line 32: Simple runner to run e2e tests for Shortest source code. Most specs are broken at this moment, will look into fixing them in a separate PR.

tsconfig.json
```diff
@@ -36,6 +36,7 @@
     ".next/types/**/*.ts"
   ],
   "exclude": [
-    "node_modules"
+    "node_modules",
+    "packages"
```
rmarescu, line 40: Both Next.js and Shortest use `@` as path alias, which causes a conflict. For that reason, I had to split the typecheck. I think Next.js should be moved to `apps/nextjs` for a better monorepo organization.

Issue #287, opened 2025-01-23
Improve caching


rmarescu:
### What

* [x] Merge CLI options into config, to simplify usage of `noCache` across the codebase
* [x] Add new config parameter `config.caching.enabled`
  * optional, default to `true`
  * overwritten by CLI arg `--no-cache`
* [x] Create one cache file per test `<TIMESTAMP>—<SHA-1>.json`
  * Having a timestamp associated should make it easier for the developer to find a test, and to purge cache chronologically
  * Use the first 8 characters of the SHA-1 value
* [x] Consider using cache versioning prefix for breaking changes
  * With certain versions the cache may break, so this would allow to migrate easily, multiple versions can coexist, easy rollback, no manual cache cleanup needed between versions 
* [x] Create a folder with the same name of the cache file to store the screenshots
  ```
  .shortest/cache/2025-01-07T04-40-01-311Z-a79d4f5a.json
  .shortest/cache/2025-01-07T04-40-01-311Z-a79d4f5a/*.png
  ```
* [ ] Add support for prompt caching for system if possible (https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching)

### Why

* In preparation for running tests in parallel, we need to use separate cache files for each test, to avoid locking conflicts
* With caching, it's difficult to tell which screenshots are associated with a given test. The screenshots should be somehow associated with the cached test.
* All cached tests are stored in a single file `cache.json`. With a project that has many tests, I think that could cause problems. Also, is hard to find and debug a given test.

rmarescu:
Moving to the next milestone to complete the remaining items.

PR #305, opened 2025-02-03
fix(cli): Component UI elements are different, running test in normal mode

### LINKED ISSUE #285: Test Execution Error: Component UI elements are different, running test in normal mode
### Version

0.4.2

### Steps to reproduce

Some tests work only once. and when returning them from the cache, they fail with the message "Test Execution Error: Component UI elements are different, running test in normal mode". I would expect the test to rerun without a cache after that error, but this is not happening. 
Test to reproduce the problem:
`shortest.beforeEach(async ({ page }) => {
    await page.goto('https://www.hitta.se/');
});

shortest('Open map, search for Stockholm, zoom in')
.expect('expect poi with houses for sale to be visible');
`

### Expected behavior

Next test run is passing - same as for the first time

### Actual behavior

Test is failing

### Additional context

_No response_

### System Info

```shell
System:
MacOs: 14.6.1 
Node version: 20.11.1
```

### Logs

_No response_

### Validations

- [x] Read the [contributing guidelines](https://github.com/anti-work/shortest/blob/main/packages/shortest/CONTRIBUTING.md).
- [x] Check that there isn't [already an issue](https://github.com/anti-work/shortest/issues) that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [discussion](https://github.com/anti-work/shortest/discussions).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

gladyshcodes:
**Fixes #285**

**What**
- Add Mutation Observer to make sure any pop-ups are loaded before taking the next step
- Move `pendingCache` into high-level function to preserve state if browser context is lost due to navigation (apparently happens when explicit timeout exceeds and page is not yet fully loaded)


packages/shortest/src/browser/core/browser-tool.ts
```diff
@@ -875,4 +875,84 @@ export class BrowserTool extends BaseBrowserTool {
...
+      });
+    } catch (error) {
+      console.log("Failed to wait for stable DOM:", error);
+      throw error;
+    }
+  }
+
+  /**
+   * DOM is considered loaded when DOMContentLoaded is fired
+   * @see https://developer.mozilla.org/en-US/docs/Web/API/Document/DOMContentLoaded_event
```
rmarescu, line 922: Shouldn't be [`@link`](https://jsdoc.app/tags-inline-link)?

```suggestion
   * @link https://developer.mozilla.org/en-US/docs/Web/API/Document/DOMContentLoaded_event
```

packages/shortest/src/ai/client.ts
```diff
@@ -42,6 +42,9 @@ export class AIClient {
     tokenUsage: { input: number; output: number };
     pendingCache: any;
   }> {
+    // temp cache store
+    const pendingCache: PendingCache = {};
```
rmarescu, line 46: If a comments is needed to document it, then should use a more appropriate name that self documents.

```suggestion
    const tempCache: TempCache = {};
```

packages/shortest/src/browser/core/browser-tool.ts
```diff
@@ -875,4 +875,84 @@ export class BrowserTool extends BaseBrowserTool {
       },
     );
   }
+
+  /**
+   * Waits for DOM to stabilize
+   * DOM is considered stabilized when:
+   * - DOMContentLoaded is fired
+   * - No mutations are detected for 1 second (e.g new elements such as modals, popups, etc.)
```
rmarescu, line 883: What if new elements are shown after more than 1 second? Or once the page scrolling is used?

packages/shortest/src/browser/core/browser-tool.ts
```diff
@@ -875,4 +875,84 @@ export class BrowserTool extends BaseBrowserTool {
...
+    }
+    try {
+      await this.waitForDOMContentLoaded();
+      await page.evaluate(async () => {
+        return await new Promise<void>((resolve) => {
+          const createTimeout = () => {
+            return setTimeout(() => {
+              resolve();
+              observer.disconnect();
+            }, 1000);
```
rmarescu, line 898: It looks like we use `1000` in 2 places (here and line 925). Can be extracted to a constant?

rmarescu (commented):
Added a few comments. Could you add a test for it? (this will also ensure the change will persiste when the AI client is migrated in https://github.com/anti-work/shortest/pull/316

Has the change been tested again the [reproduction repo](https://github.com/anti-work/shortest/issues/285#issuecomment-2609099596)? Can't tell from the PR description.


(CLA needs to be signed to merged this, as it's coming from a fork PR. You might need to fix local GH config so that commits are associated with your GH account.

gladyshcodes:
This issue will be resolved as part of the new cache implementation (#287). The fix involves updating the runCacheTest function so that it throws an error instead of returning one.

See this pr diff to get more insight.

packages/shortest/src/core/runner/index.ts
```diff
@@ -471,12 +467,10 @@ export class TestRunner {
...
-          return {
-            result: "fail" as const,
-            reason:
-              "Component UI elements are different, running test in normal mode",
-            tokenUsage: { input: 0, output: 0 },
-          };
+          console.log(
+            "Component UI elements are different, running test in normal mode",
+          );
+          throw new Error("Stale cache");
```
gladyshcodes, line 473: The fix is essentially this line.

packages/shortest/src/browser/core/browser-tool.ts
```diff
@@ -875,4 +875,84 @@ export class BrowserTool extends BaseBrowserTool {
...
     );
   }
+
+  /**
+   * Waits for DOM to stabilize
+   * DOM is considered stabilized when:
+   * - DOMContentLoaded is fired
+   * - No mutations are detected for 1 second (e.g new elements such as modals, popups, etc.)
+   */
+  public async waitForStableDOM(): Promise<void> {
```
gladyshcodes, line 885: Imo, getting rid of the explicit await new Promise((r) => setTimeout(r, MS)) should be well-thought-out, tested, and implemented as part of a separate pr

Issue #310, opened 2025-02-03
Add support for Amazon Bedrock as AI provider


rmarescu:
### What

Requires https://github.com/anti-work/shortest/issues/291

Uses https://sdk.vercel.ai/providers/ai-sdk-providers/amazon-bedrock

```javascript
export default {
  headless: false,
  baseUrl: "https://youtube.com",
  testPattern: "app/**/*.test.ts", // Don't test example tests
  ai: {
    provider: "amazon-bedrock",
    region: process.env.AWS_REGION,
    accessKeyId: process.env.AWS_ACCESS_KEY_ID,
    secretAccessKey: process.env.AWS_SECRET_ACCESS_KEY,
    model: "us.anthropic.claude-3-5-sonnet-20241022-v2:0"
  },
  mailosaur: {
    apiKey: process.env.MAILOSAUR_API_KEY,
    serverId: process.env.MAILOSAUR_SERVER_ID,
  },
} satisfies ShortestConfig;
```
* [ ] All provider props should be optional
  * [ ] Shortest code provides the [latest model available](https://sdk.vercel.ai/providers/ai-sdk-providers/amazon-bedrock#model-capabilities) (e.g. `us.anthropic.claude-3-5-sonnet-20241022-v2:0`)
  * [ ] Whitelist models that work with Shortest framework

* [ ] Automatically support ENVs
```
AWS_ACCESS_KEY_ID
AWS_SECRET_ACCESS_KEY
AWS_REGION
```
or with `SHORTEST_*` prefix
```
SHORTEST_AWS_ACCESS_KEY_ID
SHORTEST_AWS_SECRET_ACCESS_KEY
SHORTEST_AWS_REGION
```

### Why

Allows users to use Anthropic via AWS. Opens up Shortest to other models beyond Anthropic.

### Example

Bare-minimum config for the user:

```javascript
export default {
  headless: false,
  baseUrl: "https://youtube.com",
  testPattern: "app/**/*.test.ts", // Don't test example tests
  ai: {
    provider: "amazon-bedrock",
  },
  mailosaur: {
    apiKey: process.env.MAILOSAUR_API_KEY,
    serverId: process.env.MAILOSAUR_SERVER_ID,
  },
} satisfies ShortestConfig;
```

NadaaFarook:
@rmarescu I wanna take this issue

rmarescu:
[...]

Issue #291, opened 2025-01-25
Migrate to Vercel AI SDK


rmarescu:
### What

* [ ] Replace `@anthropic-ai/sdk` with [`@ai-sdk/anthropic`](https://www.npmjs.com/package/@ai-sdk/anthropic)
* [ ] New config structure (when ai.provider is `anthropic`, `ai.apiKey` is required); future provider will be `amazon-bedrock` with different props
    ```javascript
      export default {
        headless: false,
        baseUrl: "https://youtube.com",
        testPattern: "app/**/*.test.ts", // Don't test example tests
        ai: {
          provider: "anthropic",
          apiKey: process.env.ANTHROPIC_API_KEY,
          model: "claude-3-5-sonnet-20241022"
        },
        mailosaur: {
          apiKey: process.env.MAILOSAUR_API_KEY,
          serverId: process.env.MAILOSAUR_SERVER_ID,
        },
      } satisfies ShortestConfig;
    ```

* [ ] All provider props should be optional
  * [ ] Shortest code provides the [latest model available](https://docs.anthropic.com/en/docs/build-with-claude/computer-use) (e.g. `claude-3-5-sonnet-20241022`)
  * [ ] [to-be-considered] Whitelist models that work with Shortest framework?
* [ ] Automatically support ENV for the API key
    ```
    ANTHROPIC_API_KEY
    ```

    or with SHORTEST_* prefix

    ```
    SHORTEST_ANTHROPIC_API_KEY
    ```

> [!NOTE]
> All AI-related ENVs supported by Vercel AI SDK should be automatically generally supported as-si (same value), or with the `SHORTEST_` prefix (in case users would want to use separate keys)
* [ ] backward compatibity for `config.anthropicKey` with warn message

**Resources**

* https://sdk.vercel.ai/docs/ai-sdk-core
* https://sdk.vercel.ai/docs/getting-started/nodejs
* https://sdk.vercel.ai/providers/ai-sdk-providers/anthropic

### Why

Simplify the introduction of new providers and models in the future. Specifically, [Amazon Bedrock](https://github.com/anti-work/shortest/pull/268) and [OpenAI](https://openai.com/index/computer-using-agent/) once they release computer use on the API.

rmarescu:
@gladyshcodes I've updated the issue with more info based on what I learned from the investigation into https://github.com/anti-work/shortest/pull/268.

PR #316, opened 2025-02-04
feat(ai): introduce vercel ai sdk support

gladyshcodes:
**Issue #291** 
### What
Integrate [Vercel AI SDK](https://sdk.vercel.ai/docs/introduction)

### Why
Simplify the introduction of new providers and models in the future, specifically, Amazon Bedrock (#310) and [OpenAI](https://openai.com/index/computer-using-agent/) when available


gladyshcodes:
In progress:

1. Nits
2. Testing
3. Feedback iterations

packages/shortest/src/ai/client-provider.ts
```diff
@@ -0,0 +1,140 @@
...
+ */
+export function createAIClient(config: LLMConfig): LanguageModelV1 {
+  const { provider, region, accessKeyId, secretAccessKey } = config;
+  const model = resolveWhitelistedProviderModel(provider, config.model);
+
+  switch (provider) {
+    case "anthropic":
+      return createAnthropicClient(model);
+
+    case "amazon-bedrock":
```
rmarescu, line 22: Should keep this PR atomic, focused on Anthropic. AmazonBedrock can be implemented in a follow-up PR, see https://github.com/anti-work/shortest/issues/310

packages/shortest/src/ai/client-provider.ts
```diff
@@ -0,0 +1,140 @@
...
+      return createAmazonBedrockClient({
+        region,
+        accessKeyId,
+        secretAccessKey,
+        model,
+      });
+
+    default:
```
rmarescu, line 30: Is this needed? Wouldn't `parseConfig` handle invalid values?

packages/shortest/src/ai/client-provider.ts
```diff
@@ -0,0 +1,140 @@
...
+
+/**
+ * Resolves and creates the Anthropic client.
+ *
+ * @param model - The provider-specific model string.
+ * @returns A LanguageModelV1 instance for Anthropic.
+ * @throws Will throw an error if the API key is missing.
+ */
+function createAnthropicClient(model: string): LanguageModelV1 {
+  const apiKey = process.env.ANTHROPIC_API_KEY;
```
rmarescu, line 43: How would this work if the `apiKey` is coming from the config?

Generally, I think `parseConfig` (or something similar) should:
* parse the original config data
* apply ENVs as default if needed
* validate config

That way, `globalConfig` becomes the source of truth for all config-related stuff.

packages/shortest/src/ai/client-provider.ts
```diff
@@ -0,0 +1,140 @@
...
+  assert(
+    apiKey,
+    "Missing Anthropic API key. Please set process.env.ANTHROPIC_API_KEY.",
```
rmarescu, line 46: This should be part of the Zod schema, how it was before.

packages/shortest/src/ai/client.ts
rmarescu, line 1: Before removing this, we should fix all e2e specs, have a green pass, refactor, tests should still pass.

packages/shortest/src/ai/client_v2.ts
rmarescu, line 1: Why `_v2`? Will there be a `v1` around?

rmarescu (commented):
Added some initial comments.

gladyshcodes:
@rmarescu Feedback addressed

packages/shortest/src/types/config.ts
```diff
@@ -1,32 +1,119 @@
...
+  apiKey: string;
+  serverId: string;
+}
+
+/**
+ * Schema for the LLM (AI) configuration.
+ * If no model is provided, falls back to a default based on the provider.
+ */
+const llmSchema = z.object({
+  provider: z.literal(LLMSupportedProviders.ANTHROPIC),
```
rmarescu, line 31: Can this be simplified to something like:


```suggestion
  provider: z.enum(AI_PROVIDERS).default("anthropic"),
```

where

```javascript
export const AI_PROVIDERS = ["anthropic"] as const;
export type AIProvider = (typeof AI_PROVIDERS)[number];
```

packages/shortest/src/types/config.ts
```diff
@@ -1,32 +1,119 @@
...
+export interface MailosaurConfig {
+  apiKey: string;
+  serverId: string;
+}
+
+/**
+ * Schema for the LLM (AI) configuration.
+ * If no model is provided, falls back to a default based on the provider.
+ */
+const llmSchema = z.object({
```
rmarescu, line 30: Should use AI instead of LLM within our codebase.

```suggestion
const AISchema = z.object({
```

packages/shortest/src/types/config.ts
```diff
@@ -1,32 +1,119 @@
...
+    .default(getDefaultProviderModel(LLMSupportedProviders.ANTHROPIC)),
+});
+export interface LLMConfig {
+  provider: LLMSupportedProviders;
+  apiKey: string;
+  model: LLMSupportedModels;
+}
+
+export interface LLMPPublicConfig {
+  provider: LLMSupportedProvidersType;
```
rmarescu, line 44: I think even this one can be optional. Ideally, there should be zero config to run Shortest.

No `ai` prop provided should default to the provider we think is the best, e.g. `anthropic`, which tries to read the `apiKey` from ENV (as Vercel AI SDK supports it by default).

packages/shortest/src/types/config.ts
```diff
@@ -1,32 +1,119 @@
...
+ * - Validates that either `ai` or the legacy `anthropicKey` is provided (but not both).
+ * - Transforms legacy configuration to the new format.
+ */
+export const configSchema = rawConfigSchema
+  .superRefine((config, ctx) => {
+    if (!config.ai && !config.anthropicKey && !process.env.ANTHROPIC_API_KEY) {
+      ctx.addIssue({
+        code: z.ZodIssueCode.custom,
+        message:
+          "No AI configuration provided. Please provide either 'ai' or the legacy 'anthropicKey'.",
```
rmarescu, line 95: A legacy (a better terminology is _deprecated_) props should never be suggested. If is not present, then we don't want to encourage users to set it.

packages/shortest/src/types/config.ts
```diff
@@ -1,32 +1,119 @@
...
+      });
+    } else if (config.ai && config.anthropicKey) {
+      ctx.addIssue({
+        code: z.ZodIssueCode.custom,
+        message:
+          "Both 'ai' and legacy 'anthropicKey' are provided. Please use only one.",
```
rmarescu, line 101: Only the deprecated prop should be recommended for removal.

packages/shortest/src/types/config.ts
```diff
@@ -1,32 +1,119 @@
...
+      });
+    }
   })
-  .refine((config) => config.anthropicKey || process.env.ANTHROPIC_API_KEY, {
-    message:
-      "anthropicKey must be provided in config or ANTHROPIC_API_KEY environment variable",
+  .transform((config) => {
+    if (!config.ai && (config.anthropicKey || process.env.ANTHROPIC_API_KEY)) {
+      console.warn(
+        "Legacy configuration detected: the 'anthropicKey' option is deprecated and will be removed in future releases.",
```
rmarescu, line 108:
```suggestion
        "'anthropicKey' option id deprecated. Use the new 'ai' option structure instead.",
```

rmarescu (commented):
Submitting a partial review. High-level thoughts:
* No need to map Vercel's models with a local list, it complicates the implementation without too much gain
* Use `AI` instead of `LLM` (don't need 2 names that represent almost the same thing within the codebase)
* Config schema seems complicated. Can be simplified?

packages/shortest/src/types/ai.ts
```diff
@@ -1,27 +1,22 @@
...
+export enum LLMSupportedProviders {
+  ANTHROPIC = "anthropic",
 }
+export type LLMSupportedProvidersType = `${LLMSupportedProviders}`;
 
-export interface AIResponse {
-  result: "pass" | "fail";
-  reason: string;
+export enum LLMSupportedModels {
+  CLAUDE_3_5_SONNET = "claude-3-5-sonnet",
```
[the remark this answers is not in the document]
gladyshcodes: Vercel AI SDK model id differs from provider to provider. For instance:

Amazon Bedrock: anthropic.claude-3-5-sonnet-20241022-v2:0
Anthropic: claude-3-5-sonnet-20241022

packages/shortest/src/types/config.ts
gladyshcodes: I believe this approach oversimplifies the issue. Could you explain its benefits?

Right now, only the model field is optional—which makes sense because we don’t want users to worry about specifying a model. In contrast, the provider and `apiKey` fields in the configuration are mandatory

With your proposal, users would need to figure out—perhaps by reading the documentation—that the default provider is set to Anthropics. This could lead to errors if someone sets an OpenAI key in the environment but doesn’t provide the corresponding configuration

In my view, it’s much clearer when everything is explicitly defined in the configuration file. Ultimately, most users will end up setting the `ai.provider` field anyway, since only one default provider can be active even if many are available

rmarescu:
Noticed some changes from the last review, though it looks like there might still be a few things to address. Unsure if this is ready for another look. If so, just drop a comment or hit the re-review button!

rmarescu:
Remaining work (will update as needed):

* [ ] Simplify config
* [ ] Add more logging as needed
* [ ] Verify token usage calculation
* [ ] Test coverate
* [ ] Update `shortest.config.ts.example`
* [ ] Add support for `SHORTEST_ANTHROPIC_API_KEY`

rmarescu:
- [ ] Fix error when running `examples/youtube.test.rb`

<details>
<summary>Details</summary>

```
trace  | 11:23:13 | examples/youtube.test.ts | Visit a YouTube channel and verify latest content | 🤖 | 1 | Generating text | currentPrompt=
  Test: "Visit a YouTube channel and verify latest content"
  Context: {"channelName":"TED","sortBy":"newest"}
  Callback function:  [NO_CALLBACK]

  Expect:
  1. "Visit a YouTube channel and verify latest content" expected to be successful

  Current Page State:
  URL: http://localhost:3000/
  Title: Shortest
 messageCount=1 tools={
  "0": "computer",
  "1": "bash",
  "2": "github_login",
  "3": "check_email",
  "4": "sleep",
  "5": "run_callback",
  "6": "navigate"
}

error  | 11:23:16 | examples/youtube.test.ts | Visit a YouTube channel and verify latest content | 🤖 | 1 | Error making request | message=
  Invalid arguments for tool navigate: Type validation failed: Value: {"url":"https://www.youtube.com/@TED"}.
  Error message: [
    {
      "code": "invalid_literal",
      "expected": "navigate",
      "path": [
        "action"
      ],
      "message": "Invalid literal value, expected "navigate""
    }
  ]
 name=AI_InvalidToolArgumentsError stack=
  Error message: [
    {
      "code": "invalid_literal",

error  | 11:23:16 | examples/youtube.test.ts | Visit a YouTube channel and verify latest content | 🤖 | 1 | Invalid arguments for a tool were provided
error  | 11:23:16 | examples/youtube.test.ts | Visit a YouTube channel and verify latest content | 🤖 | Action failed | message=
  Invalid arguments for tool navigate: Type validation failed: Value: {"url":"https://www.youtube.com/@TED"}.
  Error message: [
    {
      "code": "invalid_literal",
      "expected": "navigate",
      "path": [
        "action"
      ],
      "message": "Invalid literal value, expected "navigate""
    }
  ]
 name=AI_InvalidToolArgumentsError stack=
  Error message: [
    {
      "code": "invalid_literal",

trace  | 11:23:16 | examples/youtube.test.ts | Visit a YouTube channel and verify latest content | 🤖 | Retry attempt | retries=1 maxRetries=3
```

</details>

packages/shortest/src/ai/prompts/index.ts
```diff
@@ -20,6 +22,7 @@ IMPORTANT GLOBAL RULES:
    - After invoking a tool, wait until the tool finishes its execution and you receive a success/failure result.
    - You will also receive metadata about the tool's execution to help you interpret its outcome.
    - Only after the tool finishes and you know the result should you request any screenshots or proceed to the next action.
+   - Always include the "action" field matching the tool name in your tool calls (e.g. for "navigate" tool, include 'action: "navigate"').
```
rmarescu, line 25: This is critial to ensure that Vercel AI SDK doesn't return `AI_InvalidToolArgumentsError`.

PR #325, opened 2025-02-07
feat(eslint): no-console-usage plugin

gladyshcodes:
### What
ESLint plugin that warns against the use of console and encourages the use of internal logging

### Why
Encourage better logging practices and ensure consistency


rmarescu:
Could this be introduced via [`no-console`](https://eslint.org/docs/latest/rules/no-console) rule instead?

Issue #281, opened 2025-01-19
Execute one test from a test file


rmarescu:
### What

When running `shortest` for a given test file, have the ability to execute a given test by passing a line number.

* works with chained tests
* callbacks work as expected

### Why

Useful to debug a given test within a file, without executing all tests. Save time & execution cost.

### Example

```shell
pnpm shortest pnpm shortest examples/youtube.test.ts:6
```

executes only [this test](https://github.com/anti-work/shortest/blob/44c67a928a814f5ec67d3f0670d5a98ec2c9d8f9/examples/youtube.test.ts#L5-L12). Line number 6 can be any value 5-12.

khalatevarun:
I wanted to confirm:

* Are we expected to use a parser (e.g., a basic tokenizer or AST-based approach) to extract function boundaries?
* would modifying the registry to store the startLine and endLine of each test be the right approach?

Any guidance on the preferred solution would be greatly appreciated! Thanks!

PR #338, opened 2025-02-12
feat(cli): execute one test from a test file 

khalatevarun:
fix: #281 

Affected regions:
* `runTests()`
* `executeTestFile()`
* `main()`
* package.json - new packages added to parse the file - `acorn` and `acorn-walk`


rmarescu (commented):
I'll be able to take a closer look on this during the next milestone (next week or so). Generally, I think it should use a parser like `@babel/parser` as it is more reliable than regex.

Here is a [sample AI implementation](https://claude.site/artifacts/f0705725-9487-41ab-8665-251ffa1f8da9), for reference/inspiration (haven't looked into details nor tested the code)

khalatevarun:
@rmarescu I request you to please take a look at this PR as its ready and synced with latest development
@gladyshcodes would like to know your thoughts on this too

rmarescu (commented):
It appears that Babel would be a better option for us.

- **TypeScript/JSX Support**: Since Shortest is written in TypeScript, Babel's native TypeScript parsing will more accurately handle test files.
- **Parent-Child Relationship Tracking**: The parentPath feature in Babel's traversal is ideal for following method chains - essential for our use case where `shortest()` calls have chained `expect()` methods.
- **Scope Awareness**: Babel can identify related test assertions even when they span multiple lines or appear in different expressions but the same scope.
- **Visitor Pattern**: Babel's visitor pattern is more efficient than current nested AST walks.
- **Type Safety**: Better TypeScript types for AST nodes will make your code more maintainable.

Although Acorn generally has better performance metrics:
- The performance difference is minimal for our specific use case
- We're parsing individual test files, not entire codebases
- The improved accuracy in test detection far outweighs any slight performance difference

For example, the current logic doesn't parse the updated `api-assert-bearer.test.ts` correctly:

```
Test Discovery: No tests found at line 7 in examples/api-assert-bearer.test.ts
```
CleanShot 2025-02-27 at 10 37 24@2x

- - -
Will work on replacing with Babel in this PR, and will add `TestCase` in a separate one.

PR #357, opened 2025-02-20
feat(eslint): enforce Zod schema types

rmarescu:
Add a custom ESLint rule that enforces using Zod schemas for runtime type validation.

## What
- Add `require-zod-schema-types` ESLint rule
- Configure rule to exempt React and Drizzle ORM types
- Add `eslint-disable` comments for existing type-only definitions
- Include vitest.config.ts in TypeScript project

## Why
- Ensures consistent runtime type validation by requiring Zod schemas
- Enforces new code moving forward uses Zod schema
- Prevents type-only types that lack runtime validation

## Future work
- Convert existing type-only definitions to Zod schemas
- Extend rule to enforce Zod over interfaces

### DIFF
--- a/app/(dashboard)/dashboard/types.ts
+++ b/app/(dashboard)/dashboard/types.ts
@@ -21,6 +21,7 @@ export interface TestFile {
   oldContent?: string;
 }
 
+// eslint-disable-next-line zod/require-zod-schema-types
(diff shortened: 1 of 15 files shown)


rmarescu (commented):
This may be worth adding to the other projects.

If it works well, we could extract and open source as a stand-alone package.

PR #360, opened 2025-02-21
refactor: reorganize cache implementation

rmarescu:
- Move `BaseCache` to dedicated file
- Create specialized `TestCache` class
- Add `cache:clear` script command

### Why
* Part of #287
* Preparation to implement per-test-cache storage


packages/shortest/src/cache/base-cache.ts
rmarescu, line 1: Moved as-is from `./index.ts`

packages/shortest/src/core/runner/index.ts
```diff
@@ -457,6 +456,10 @@ export class TestRunner {
             await browserTool.getNormalizedComponentStringByCoords(x, y);
 
           if (componentStr !== step.extras.componentStr) {
+            this.log.trace("Component UI elements mismatch", {
```
rmarescu, line 459: Noticed that the `Log in` test cannot use the cache version, possible related to #285.
Will investigate separately.

PR #362, opened 2025-02-23
feat: make config.testPattern optional

rmarescu:
* Make `config.testPattern` optional, default to `**/*.test.ts`
* Improve test runner error handling
* Improve CLI help output formatting and organization

### Why

* One step closer towards zero-config
* Reduce complexity and technical debt around error handling and default values

CleanShot 2025-02-22 at 16 26 13@2x

### DIFF
--- a/packages/shortest/src/cli/bin.ts
+++ b/packages/shortest/src/cli/bin.ts
@@ -40,48 +40,37 @@ ${pc.bold("Shortest")} - AI-powered end-to-end testing framework
 ${pc.dim("https://github.com/anti-work/shortest")}
 
 ${pc.bold("Usage:")}
-  shortest [options] [test-pattern]
-  shortest cache clear [--force-purge]
-
-${pc.bold("Commands:")}
-  init                  Initialize Shortest in the current directory
-  cache clear           Clear test cache
-    --force-purge       Force delete all cache files
+  $ shortest [options] [pattern]                  Run tests matching pattern (default: **/*.test.ts)
+  $ shortest init                                 Initialize Shortest in current directory
+  $ shortest cache clear [--force-purge]          Clear test cache
 
 ${pc.bold("Options:")}
-  --headless            Run tests in headless browser mode
-  --log-level=<level>   Set log level (default: silent). Options: silent, error, warn, info, debug, trace
-  --target=<url>        Set target URL for tests (default: http://localhost:3000)
-  --github-code         Generate GitHub 2FA code for authentication
-  --no-cache            Disable caching (storing browser actions between tests)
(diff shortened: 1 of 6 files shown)


packages/shortest/src/cli/bin.ts
```diff
@@ -205,7 +194,7 @@ const main = async () => {
     log.trace("Initializing TestRunner");
     const runner = new TestRunner(process.cwd(), config);
     await runner.initialize();
-    const success = await runner.runTests(testPattern);
+    const success = await runner.execute(config.testPattern);
```
rmarescu, line 197: This name better represents what the function does (at the time of calling, is not known yet if there are any tests to be executed yet).

packages/shortest/src/core/runner/index.ts
```diff
@@ -61,29 +60,6 @@ export class TestRunner {
...
-    if (files.length === 0) {
-      this.reporter.error(
-        "Test Discovery",
-        `No test files found matching: ${testPattern}`,
-      );
-      this.log.error("No test files found matching", {
-        pattern: testPattern,
-      });
-      process.exit(1);
-    }
```
rmarescu, line 82: This is duplicated started on line 375 below.

packages/shortest/src/core/runner/index.ts
```diff
@@ -61,29 +60,6 @@ export class TestRunner {
     this.browserManager = new BrowserManager(this.config);
   }
 
-  private async findTestFiles(pattern?: string): Promise<string[]> {
-    this.log.trace("Finding test files", { pattern });
-    const testPattern = pattern || this.config.testPattern || "**/*.test.ts";
```
rmarescu, line 66: Config should be the source of truth of all default values, in conjuction with Zod validation.

packages/shortest/src/core/runner/index.ts
```diff
@@ -61,29 +60,6 @@ export class TestRunner {
     this.browserManager = new BrowserManager(this.config);
   }
 
-  private async findTestFiles(pattern?: string): Promise<string[]> {
```
rmarescu, line 64: Once the default value assignment and duplicated logic are removed, there isn't much value having this function.

packages/shortest/tests/unit/config.test.ts
```diff
@@ -7,22 +7,9 @@ import { parseConfig } from "@/utils/config";
 describe("Config parsing", () => {
   let baseConfig: ShortestConfig;
 
-  beforeEach(() => {
```
rmarescu, line 10: Moved below (not used by `with minimal config` test)

PR #363, opened 2025-02-23
fix: improve test cache execution with fallback to normal mode

rmarescu:
* Bump Vercel AI SDK packages for both Next.js and Shortest
* Improve click animation handling with proper timing
* Add better error handling with typed cache errors
* Fix cursor position tracking in browser tool
* Clean up test cache file deletion logic
* Add trace logging for test start/end events
* Refactor error types using Zod enums

### Why

* Attempt to fix https://github.com/anti-work/shortest/issues/285
* While the test run now gracefully falls back to executing the test in normal mode, the actual caching mismatch may still be an issue that requires a deeper (reproducible on our `Log in` test for the Next.js app)
* Reduce coupling between `executeTest` and `runCachedTest` by encapsulating all cache-related logic in `runCachedTest`

### DIFF
--- a/packages/shortest/src/utils/errors.ts
+++ b/packages/shortest/src/utils/errors.ts
@@ -9,7 +9,6 @@ export class ShortestError extends Error {
   }
 }
 
-// eslint-disable-next-line zod/require-zod-schema-types
 const ConfigErrorTypeSchema = z.enum([
   "duplicate-config",
   "file-not-found",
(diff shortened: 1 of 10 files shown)


packages/shortest/src/utils/errors.ts
```diff
@@ -1,43 +1,67 @@
...
+import { z, ZodError } from "zod";
 
-// eslint-disable-next-line zod/require-zod-schema-types
-export type ConfigErrorType =
-  | "duplicate-config"
-  | "file-not-found"
-  | "invalid-config"
-  | "no-config";
-export class ConfigError extends Error {
+export class ShortestError extends Error {
```
rmarescu, line 3: Adding our own error class so that we can filters the errors that need to be caught. Now we do a catch-all, which is not great.

PR #369, opened 2025-02-28
feat: add claude-3-7-sonnet-20250219

rmarescu:
* [x] Add 2 new model options for config.ai.model for `anthropic`
  * `claude-3-7-sonnet-20250219 `
  * `claude-3-7-sonnet-latest `
* [x] Add 2 new computer tools: `computer_20250124`, `bash_20250124`
* [x] Add support for new actions: `triple_click`, `hold_key`, `left_mouse_down`, `left_mouse_up`, `wait`, `scroll`

> [!NOTE]
> This PR does not add support for [reasoning](https://sdk.vercel.ai/providers/ai-sdk-providers/anthropic#reasoning)

PR #372, opened 2025-03-02
refactor: add ToolRegistry

rmarescu:
### What

Implement a comprehensive refactoring of the AI tools system  

* Create a modular tool system with a central ToolRegistry class
* Move tools from hardcoded implementation in AIClient to dedicated files
* Implement tool registration through createToolRegistry()
* Map Anthropic's computer actions to internal actions

### Why

In preparation for adding Claude 3.7.

Ref https://github.com/anti-work/shortest/pull/369#issuecomment-2691388346

### DIFF
--- a/packages/shortest/src/tools/tool-registry.ts
+++ b/packages/shortest/src/tools/tool-registry.ts
@@ -4,14 +4,15 @@ import { BrowserTool } from "@/browser/core/browser-tool";
 import { getLogger, Log } from "@/log";
 import { AnthropicModel } from "@/types/config";
 import { ShortestError } from "@/utils/errors";
+
+const TOOL_ENTRY_CATEGORIES = ["provider", "custom"] as const;
+const toolEntryCategorySchema = z.enum(TOOL_ENTRY_CATEGORIES);
+const toolFactoryNoArgSchema = z.function().args().returns(z.custom<Tool>());
+
 const toolFactoryWithArgSchema = z
   .function()
   .args(z.custom<BrowserTool>())
   .returns(z.custom<Tool>());
-
-export const TOOL_ENTRY_CATEGORIES = ["provider", "custom"] as const;
-const toolEntryCategorySchema = z.enum(TOOL_ENTRY_CATEGORIES);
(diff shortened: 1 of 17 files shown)


packages/shortest/src/tools/tool-registry.ts
```diff
@@ -0,0 +1,164 @@
...
+  {
+    "claude-3-5-sonnet-latest": "claude-3-5",
+    "claude-3-5-sonnet-20241022": "claude-3-5",
+  };
+
+const ANTHROPIC_TOOL_VERSION_MAP: Record<
+  AnthropicModelFamily,
+  Record<AnthropicToolType, AnthropicToolVersion>
+> = {
+  "claude-3-5": {
```
rmarescu, line 54: Easy to extend with 3.7.

packages/shortest/src/tools/tool-registry.ts
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    try {
+      const bashToolEntry = this.getProviderToolEntry(provider, model, "bash");
+      // @ts-ignore
+      // For some reason, it expects an argument, but it doesn't take any
+      tools["bash"] = bashToolEntry.factory();
+    } catch (error) {
+      if (!(error instanceof ShortestError)) throw error;
+      this.log.trace("Bash tool not found for model, skipping", { model });
```
rmarescu, line 167: If a tool is not found (e.g. bash for 3.7 not yet available upstream), the package will still work without the tool.

packages/shortest/tests/unit/tools/index.test.ts
```diff
@@ -0,0 +1,128 @@
...
+      );
+
+      Object.values(tools).forEach((tool) => {
+        expect(tool).toHaveProperty("execute");
+
+        expect(typeof tool.execute).toBe("function");
+      });
+    });
+
+    it("registers tools with expected keys", () => {
```
rmarescu, line 112: This ensures all the tools available are properly registered. If a tool doesn't get registered, and not passed to the AI, the AI will do its best to complete the task with the missing tool. It will not be obvious - through a script error - that a tool is missing).

PR #373, opened 2025-03-02
feat: add config.browser options

rmarescu:
Adds browser.contextOptions to pass custom [Playwright browser context options](https://playwright.dev/docs/test-use-options#more-browser-and-context-options).
* Allows configuring browser behavior (ignore HTTPS errors, permissions, geolocation)
* Apply options when initializing browser context

### DIFF
--- a/packages/shortest/src/types/config.ts
+++ b/packages/shortest/src/types/config.ts
@@ -1,3 +1,4 @@
+import { BrowserContextOptions } from "playwright";
 import { z } from "zod";
 
 export const cliOptionsSchema = z.object({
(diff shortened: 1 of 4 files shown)


packages/shortest/src/types/config.ts
```diff
@@ -41,10 +42,18 @@ const mailosaurSchema = z
...
+  /**
+   * @see https://playwright.dev/docs/api/class-browser#browser-new-context
+   */
+  contextOptions: z.custom<BrowserContextOptions>().optional(),
+});
+
 export const configSchema = z
   .object({
     headless: z.boolean().default(true),
     baseUrl: z.string().url("must be a valid URL"),
```
rmarescu, line 55: This should be moved into `browserSchema`, and ideally made optional (with default `about:blank`).

rmarescu (commented):
Tested on Flexile CI.

Issue #379, opened 2025-03-06
shortest init command skips steps if package already installed


khalatevarun:
### Version

0.4.5

### Your repository URL

https://github.com/khalatevarun/shortest-example

### Actual behavior

1. Created a fresh new nextjs project and installed @antiwork/shortest using npm i @antiwork/shortest
2. Run npx @antiwork/shortest init to setup rest of the configs
3. The cli returned the following and process terminated
```
Setting up Shortest...
✔ Package already installed
```

### Expected behavior

The cli should continue with the rest of the steps if the package is already installed
npx @antiwork/shortest init
```
Setting up Shortest...
✔ Package already installed
✔ shortest.config.ts created
✔ .env.local created (ANTHROPIC_API_KEY added)
✔ .gitignore created

Initialization complete! Next steps:
1. Update .env.local with your values
2. Create your first test file: example.test.ts
3. Run tests with: npx/pnpm test example.test.ts
```

### System Info

```shell
System:
    OS: macOS 14.1
    CPU: (11) arm64 Apple M3 Pro
    Memory: 111.08 MB / 18.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.17.0 - ~/.nvm/versions/node/v20.17.0/bin/node
    npm: 10.8.2 - ~/.nvm/versions/node/v20.17.0/bin/npm
    pnpm: 9.15.1 - ~/.nvm/versions/node/v20.17.0/bin/pnpm
  Browsers:
    Chrome: 134.0.6998.44
    Safari: 17.1
```

### Validations

- [x] I've checked for [existing issues](https://github.com/anti-work/shortest/issues)
- [x] This is a concrete bug (for Q&A use [discussions](https://github.com/anti-work/shortest/discussions))

PR #297, opened 2025-01-27
feat(cli): Add `shortest init` command

khalatevarun:
Resolves #294 

CleanShot 2025-01-29 at 23 31 36@2x

PR #380, opened 2025-03-06
fix(cli): shortest init command

khalatevarun:
fix: #379

### DIFF
--- a/packages/shortest/src/cli/init/index.ts
+++ b/packages/shortest/src/cli/init/index.ts
@@ -20,12 +20,11 @@ export default async function main() {
       packageJson?.devDependencies?.["@antiwork/shortest"]
     ) {
       console.log(pc.green("✔ Package already installed"));
-    }
-    else {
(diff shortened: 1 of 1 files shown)


packages/shortest/src/cli/init/index.ts
```diff
@@ -20,12 +20,12 @@ export default async function main() {
       packageJson?.devDependencies?.["@antiwork/shortest"]
     ) {
       console.log(pc.green("✔ Package already installed"));
-      return;
```
rmarescu, line 23: The current behaviour is intentional. If the package is already present, we can assume that some sort of setup has already been done. Otherwise, trying to do an incremental install could trigger various edge cases that are not worth supporting at this time. 

See https://github.com/anti-work/shortest/pull/297/files#r1935142270

khalatevarun: noted, thanks for pointing out

Issue #381, opened 2025-03-06
execute test with line no not working


khalatevarun:
### Version

0.4.5

### Your repository URL

https://github.com/khalatevarun/shortest-example.git

### Actual behavior

npx shortest example.test.ts:4
Found 1 test file(s)
file:///Users/varunk/playground/shortest-example/node_modules/@antiwork/shortest/dist/cli/bin.js:7931
    traverse(ast, {
    ^

TypeError: traverse is not a function
    at parseShortestTestFile (file:///Users/varunk/playground/shortest-example/node_modules/@antiwork/shortest/dist/cli/bin.js:7931:5)
    at TestRunner.filterTestsByLineNumber (file:///Users/varunk/playground/shortest-example/node_modules/@antiwork/shortest/dist/cli/bin.js:8374:27)
    at TestRunner.executeTestFile (file:///Users/varunk/playground/shortest-example/node_modules/@antiwork/shortest/dist/cli/bin.js:8415:33)
    at async TestRunner.execute (file:///Users/varunk/playground/shortest-example/node_modules/@antiwork/shortest/dist/cli/bin.js:8506:7)
    at async main2 (file:///Users/varunk/playground/shortest-example/node_modules/@antiwork/shortest/dist/cli/bin.js:8724:21)

Node.js v20.17.0

### Expected behavior

Should find and execute the test

### System Info

```shell
System:
    OS: macOS 14.1
    CPU: (11) arm64 Apple M3 Pro
    Memory: 111.08 MB / 18.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.17.0 - ~/.nvm/versions/node/v20.17.0/bin/node
    npm: 10.8.2 - ~/.nvm/versions/node/v20.17.0/bin/npm
    pnpm: 9.15.1 - ~/.nvm/versions/node/v20.17.0/bin/pnpm
  Browsers:
    Chrome: 134.0.6998.44
    Safari: 17.1
```

### Validations

- [x] I've checked for [existing issues](https://github.com/anti-work/shortest/issues)
- [x] This is a concrete bug (for Q&A use [discussions](https://github.com/anti-work/shortest/discussions))

PR #382, opened 2025-03-06
fix(cli): execute test with line no functionality

khalatevarun:
fix: #381 

@babel/traverse is a CommonJS module,
which is being imported into ESM, 
hence its exports structure becomes { default: { default: Function } }

We need to access the nested default export and use the commonjs require approach


rmarescu (approved):
I thought the feature was working fine once #338 got merged. I looked into it, and I think this last [clean-up commit](https://github.com/anti-work/shortest/pull/338/commits/642119f8c0b7d732e6419894fd57415bdc8592b2) messed it up (and probably I didn't do a final check). Thanks for the fix!

PR #385, opened 2025-03-09
feat: QA automation setup wizard

### LINKED ISSUE #384: Hosted Shortest test runner
### What

A button/wizard that spins up end-to-end QA automation for any web app.

- Login with GitHub
- Select a repository
- Select test folder/pattern
- Commit shortest.config.ts file to repo
- Commit GitHub Action that runs these tests on every commit to `main`, pull request commit (booleans in UI?)

Next:

- Automatic test suite generation (shortlist of typical user flows)
- More options around config + GitHub action workflow

### Why

We should ship this incrementally, and this would allow us to start building out a web app experience without hijacking the core shortest package. Over time, we can get smarter about generating tests and simplifying the "tests" that live within the repo. However, the tests _should_ still live in the repo as that's the right place for tests to live – within version control, per commit, per PR, etc.

The shortest hosted service will be "smarter" than just running those tests. It could look at PRs and make suggestions by running tests and committing them to the codebase upon successful run (or upon the user confirming they want those added).

### Example

_No response_

crisog:
Hey there! I'm new around here but really eager to start contributing to Shortest :)

I was looking for issues and found #384 interesting, gave it a shot. Feel free to provide feedback.

The PR is pretty big which I don't love, but since I'm not a member I believe it doesn't make sense to tackle this with incremental PRs. I could break this down in smaller PRs if you guys like what I did here.

I'm still running into issues getting the workflow to run due to caching trying to access the runner filesystem (despite having caching disabled in config) but working on that.

Closes #384

Video Demo
https://cap.so/s/hgxtq56qrchvy34


app/api/github/setup/route.ts
```diff
@@ -0,0 +1,206 @@
...
+    try {
+      // Create or update files serially to avoid conflicts
+      await createOrUpdateFile(
+        octokit,
+        owner,
+        repo,
+        "shortest.config.ts",
+        configContent,
+        "Add Shortest configuration",
+        defaultBranch,
```
rmarescu, line 163: Generally, `main` branches are protected, and I don't think Shortest should push commits directly to `main`. In a CI setup, changes to `main` may trigger a production deployment, and there is a significant risk allowing a 3rd party app to have that type of access.

app/api/github/setup/route.ts
```diff
@@ -0,0 +1,206 @@
...
+      );
+
+      await createOrUpdateFile(
```
rmarescu, line 166: Any code change that comes from Shortest setup should be tested to ensure that works as intended. What happens in this scenario if the workflow run fails?

Another scenario - what happens if the script fails on line 165 above? The repo will have a half-baked installation, from which is more difficult to recover.
The best way to reduce the complexity damage caused by exception handling is to reduce the number of places where exceptions have to be handled.

In this particular case, a design that can do that is having the code commited to a PR. From Shortest's perpective:
* if the PR build is green, then the setup was successful, and it can be merged
* if for any reason it fails mid-way, or there are other issues, we could just trash it, and try again (don't have to go into specific edge cases and try to recover from there)

So just taking a different path from the start on implementing this, it can simplify the overall setup.

app/api/github/setup/route.ts
```diff
@@ -0,0 +1,206 @@
...
+
+on:
+${triggers.onPush ? `  push:\n    branches:\n      - ${defaultBranch}\n` : ""}
+${triggers.onPullRequest ? "  pull_request:\n" : ""}
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    env:
+      ANTHROPIC_API_KEY: \${{ secrets.ANTHROPIC_API_KEY }}
```
rmarescu, line 142: This assumes the project uses this ENV variable (added some [thoughts](https://github.com/antiwork/shortest/issues/384#issuecomment-2712883594) on some initial assumptions that need to be made).

If a project uses Anthropic internall, we could assume they use `ANTHROPIC_API_KEY` env.   While Shortest can use the same ENV, the user might want to separate cost from their main API key. For that reason, Shortest supports `SHORTEST_ANTHROPIC_API_KEY`, which I think it should be the default.

app/api/github/setup/route.ts
```diff
@@ -0,0 +1,206 @@
...
+
+    const configContent = `import type { ShortestConfig } from "@antiwork/shortest";
+
+export default {
+  headless: false,
+  baseUrl: "http://localhost:3000",
+  testPattern: '${testPattern}',
+  ai: {
+    provider: "anthropic",
+    apiKey: process.env.ANTHROPIC_API_KEY
```
rmarescu, line 128:
```suggestion
```

This assumes the project uses this ENV variable (added some [thoughts](https://github.com/antiwork/shortest/issues/384#issuecomment-2712883594) on some initial assumptions that need to be made).

If a project uses Anthropic internall, we could assume they use `ANTHROPIC_API_KEY` env.   While Shortest can use the same ENV, the user might want to separate cost from their main API key. For that reason, Shortest supports `SHORTEST_ANTHROPIC_API_KEY`, which I think it should be the default.

By removing this line, it would default to that.

app/api/github/setup/route.ts
```diff
@@ -0,0 +1,206 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Node.js
+        uses: actions/setup-node@v2
+        with:
+          node-version: '22'
+      - name: Install dependencies
+        run: npm install
+      - name: Run Shortest tests
+        run: npx shortest run`;
```
rmarescu, line 152: Shortest needs a web server to run. It will try to hit `http://localhost:3000/` that is not available.

app/(dashboard)/dashboard/setup/page.tsx
```diff
@@ -0,0 +1,31 @@
...
+export default async function SetupPage() {
+  const configs = await getUserRepositoryConfigs();
+
+  return (
+    <div className="space-y-6 h-full flex flex-col">
+      <div className="p-6">
+        <div className="flex items-center justify-between">
+          <div>
+            <h1 className="text-2xl font-semibold tracking-tight">
+              QA Automations
```
rmarescu, line 14: This is more about managing / viewing your projects. Check Vercel.com UI for inspiration.

app/(dashboard)/layout.tsx
```diff
@@ -17,28 +21,65 @@ const Logo = () => (
...
+              <SignedOut>
+                <Link href="/" className="flex items-center">
+                  <Logo />
+                </Link>
+              </SignedOut>
+
+              <SignedIn>
+                <nav className="flex items-center space-x-4">
+                  <NavLink href="/dashboard">Dashboard</NavLink>
+                  <NavLink href="/dashboard/setup">Setup</NavLink>
```
rmarescu, line 70: The path for this should be `/projects` (to match the table name), and keep it simple.

rmarescu (commented):
Hey @crisog, thanks for the interest in Shortest!

I've added some high-level comments. I've tried the flow on local, and got stuck at the GH integration.

CleanShot 2025-03-11 at 08 48 00@2x

Here are [some notes](https://github.com/antiwork/shortest/issues/384#issuecomment-2712883594) I added last night about how I think this implementation should work.

app/api/github/setup/route.ts
crisog: good point! are we good opening a PR instead?

app/api/github/setup/route.ts
crisog: Based on [the assumptions](https://github.com/antiwork/shortest/issues/384#issuecomment-2712883594) on the associated issue,

Then we should ask the user for the start script of the app & the port (so we can use those on the workflow & config) - we could assume the start script, but probably not the port.

crisog: > Shortest needs a web server to run. It will try to hit `http://localhost:3000/` that is not available.

Based on [the assumptions](https://github.com/antiwork/shortest/issues/384#issuecomment-2712883594) on the associated issue,

Then we could ask the user for the start script of the app & the port (so we can use those on the workflow & config) - but that would require the Github env to have all necessary variables to run the app. I believe if someone wants to do e2e tests, that's something the should expect though.

The other option would be to simply use a hosted app URL and have this workflow only trigger manually? (so the user can control when the deployment is up and ready)

crisog: Assuming that MVP only supports Vercel, we can generate a workflow similar to this
https://github.com/antiwork/shortest/blob/main/.github/workflows/web-shortest.yml

However, we'd also need to figure out what to do with the package manager (we could have support for the most popular ones, and generate workflow with the install scripts based on the configured package manager?)

crisog: The same with other setup scripts (i.e. the drizzle scripts in the web-shortest.yml)

rmarescu:
@crisog we had an internal discussion to pause development on this for now. Closing, can reopen if needed.

PR #387, opened 2025-03-09
fix(tools): change the way how screenshot is cached

devsargam:
Part of https://github.com/anti-work/shortest/issues/287

The core change is moving from a shared `screenshots` folder to dedicated folders per test run. This improves organization and makes test results easier to manage and reference.

- Replace TestCache with TestRun and TestRunRepository system
- Organize screenshots in dedicated folders per test run
- Improve test artifact persistence with separate repository directories
- Improve versioning for test run data formats
- Implement retention policies for managing test runs


devsargam:
Currently, I have added new way of taking screenshot but also supporting the older way of doing so. If we don't want backwards compatibility please let me know!

packages/shortest/src/browser/core/browser-tool.ts
```diff
@@ -806,7 +807,33 @@ export class BrowserTool extends BaseBrowserTool {
...
+        // Save screenshot in test-specific directory
+        screenshotPath = join(testScreenshotDir, `screenshot-${timestamp}.png`);
+      } else {
+        // Fallback if identifier is undefined
+        screenshotPath = join(
+          this.screenshotDir,
+          `screenshot-${timestamp}.png`,
+        );
+      }
+    } else {
```
rmarescu, line 833: Is there a case when a screenshot is taken, and there is no `testContext`?

packages/shortest/src/browser/core/browser-tool.ts
```diff
@@ -806,7 +807,33 @@ export class BrowserTool extends BaseBrowserTool {
...
-    const filePath = join(this.screenshotDir, `screenshot-${timestamp}.png`);
+    let screenshotPath: string;
+
+    // If we have a test context with a current test that is a TestCase, use test-specific directory
+    if (
+      this.testContext?.currentTest &&
+      "identifier" in this.testContext.currentTest
+    ) {
+      const testCase = this.testContext.currentTest as TestCase;
+      if (testCase.identifier) {
```
rmarescu, line 818: A `testCase` always has an `identifier`, [generated via `transform`](https://github.com/devsargam/shortest/blob/93e57bed4762587bcfdcad7727b59257e4002b03/packages/shortest/src/core/runner/test-case.ts#L57-L62). The [schema](https://github.com/devsargam/shortest/blob/93e57bed4762587bcfdcad7727b59257e4002b03/packages/shortest/src/core/runner/test-case.ts#L44-L63) should be updated with a `pipe` to make it `strict()`

```typescript
  // ...
  .transform((data) => {
    const hashInput = `${data.name}:${data.filePath}:${JSON.stringify(data.expectations)}`;
    // Low collision risk for datasets under 65,000 tests
    data.identifier = createHash(hashInput, { length: 8 });
    return data;
  })
  .pipe(z.object({ identifier: z.string() }).strict());
```

packages/shortest/src/browser/core/browser-tool.ts
```diff
@@ -806,7 +807,33 @@ export class BrowserTool extends BaseBrowserTool {
...
+        const testScreenshotDir = join(CACHE_DIR_PATH, testCase.identifier);
```
rmarescu, line 819:
```suggestion
        const testScreenshotDir = join(CACHE_DIR_PATH, testCase.identifier);
```

If the screenshot folder is only the identifier, that means 2 different runs will use the same folder. When that happens, the cache file is removed, but the screenshot are not.

To address it, the screenshot folder should match the name of the cache file, e.g. `<timestamp>_<identifier>`. Let's add a new property on `TestCase` for `startedAt` (via `transform`, similar to `identifier`), and use that to craft the folder name, and cache file name.
When the cache file is removed, the associated folder should be removed as well. 
That way, the metadata for a test run is self-contained.

packages/shortest/src/browser/core/browser-tool.ts
```diff
@@ -879,12 +907,36 @@ export class BrowserTool extends BaseBrowserTool {
 
   private cleanupScreenshots(): void {
```
rmarescu, line 908: I don't think this is needed anymore, since no more screenshots are created within `screenshots` folder. Instead, we should clean up the legacy folder via a new function `purgeLegacyScreenshots`, similar to `purgeLegacyCache`.

packages/shortest/src/browser/core/browser-tool.ts
```diff
@@ -879,12 +907,36 @@ export class BrowserTool extends BaseBrowserTool {
...
+          this.cleanupScreenshotsInDir(testScreenshotDir);
+        }
+      }
+    } catch (error) {
+      this.log.error("Failed to clean up screenshots", getErrorDetails(error));
+    }
+  }
+
+  // Helper method to clean up screenshots in a specific directory
+  private cleanupScreenshotsInDir(directory: string): void {
```
rmarescu, line 930: Test run screenshot folder clean-up should be done as part of `cleanUpCache`

packages/shortest/src/browser/core/browser-tool.ts
```diff
@@ -806,7 +807,33 @@ export class BrowserTool extends BaseBrowserTool {
 
   private async takeScreenshotWithMetadata(): Promise<ToolResult> {
     const timestamp = new Date().toISOString().replace(/[:.]/g, "-");
-    const filePath = join(this.screenshotDir, `screenshot-${timestamp}.png`);
+    let screenshotPath: string;
+
+    // If we have a test context with a current test that is a TestCase, use test-specific directory
+    if (
+      this.testContext?.currentTest &&
```
rmarescu, line 814: Is there a case when a screenshot is taken, and there is no `testContext`?

rmarescu (commented):
Hey @devsargam, thanks for the contribution. I've added some comments.

packages/shortest/src/browser/core/browser-tool.ts
devsargam:
```ts
const TestCaseSchema = z
  .object({
    name: z.string(),
    filePath: z.string(),
    payload: z.any().optional(),
    fn: TestCaseFunctionSchema.optional(),
    expectations: z.array(TestCaseExpectationsSchema).default([]),
    beforeFn: TestCaseFunctionSchema.optional(),
    afterFn: TestCaseFunctionSchema.optional(),
    directExecution: z.boolean().optional().default(false),
    identifier: z.string().optional(),
  })
  .strict()
  .transform((data) => {
    const hashInput = `${data.name}:${data.filePath}:${JSON.stringify(data.expectations)}`;
    // Low collision risk for datasets under 65,000 tests
    data.identifier = createHash(hashInput, { length: 8 });
    return data;
  });
```

Here, regardless of whether user passes identifier or not, the original is going to be overwritten by what's there in the transform? Is this intentional?

packages/shortest/src/browser/core/browser-tool.ts
devsargam: That was for maintaining backwards compatibility, but I understand now

devsargam:
TODOS:
- [ ] figure out way to delete screenshots inside directory with MAX_SCREENSHOT_LIMIT
- [ ] test changes thoroughly

packages/shortest/src/browser/core/browser-tool.ts
```diff
@@ -415,26 +404,20 @@ export class BrowserTool extends BaseBrowserTool {
           };
 
         case InternalActionEnum.RUN_CALLBACK: {
-          if (!this.testContext?.currentTest) {
```
rmarescu, line 418: `testContext` must always be present.

packages/shortest/src/browser/core/browser-tool.ts
```diff
@@ -877,40 +867,6 @@ export class BrowserTool extends BaseBrowserTool {
     this.testContext = newContext;
   }
 
-  private cleanupScreenshots(): void {
```
rmarescu, line 880: `TestRunRepository` deals with cleaning up unwanted runs now.

packages/shortest/src/cache/index.ts
```diff
@@ -31,37 +29,65 @@ export const cleanUpCache = async ({
...
-      if (forcePurge || now - entry.metadata.timestamp > CACHE_MAX_AGE_MS) {
-        await fs.unlink(filePath);
-        log.trace("Cache file removed", { file: filePath });
+      // Check if cache is outdated or orphaned
+      const isOutdatedVersion =
+        entry.metadata.version < TestRunRepository.VERSION;
+      // TODO: Check lineNumber once that is available
+      const testFileExists = existsSync(
+        path.join(process.cwd(), entry.test.filePath),
+      );
```
rmarescu, line 67: Rudimentary check if the entry is _orphaned_.

packages/shortest/src/cache/test-cache.ts
rmarescu, line 1: Replaced mostly by TestRunRepository

packages/shortest/src/core/runner/index.ts
```diff
@@ -82,32 +89,37 @@ export class TestRunner {
...
   ): Promise<TestRun> {
+    const testCase = testRun.testCase;
     this.log.trace("Executing test", {
       name: testCase.name,
       filePath: testCase.filePath,
       payload: testCase.payload,
       skipCache,
     });
-    const testRun = new TestRun(testCase);
-    testRun.markRunning();
```
rmarescu, line 106: This is now initialized by the caller of this function.

packages/shortest/src/core/runner/index.ts
```diff
@@ -363,13 +370,40 @@ export class TestRunner {
           }
 
           this.reporter.onTestStart(testCase);
-          const testRun = await this.executeTest(testCase, context);
+          const testRun = new TestRun(testCase);
+          await testRun.initialize();
+          try {
+            testRun.markRunning();
```
rmarescu, line 374: It needs to be outside becauase `executeTest` calls itself when cached run doesn't work.

packages/shortest/src/core/runner/test-run-repository.ts
```diff
@@ -0,0 +1,411 @@
...
+    log.error("Uncaught exception", getErrorDetails(error));
+    await cleanUpAndExit();
+  });
+
+  handlersRegistered = true;
+  return activeRepositories;
+};
+
+export class TestRunRepository {
+  public static VERSION = 2;
```
rmarescu, line 64: Increased from `1` previously, and using numbers now. Generally, test runs with previous versions are automatically discarted, their metadata is different.

packages/shortest/src/core/runner/test-run-repository.ts
```diff
@@ -0,0 +1,411 @@
...
+    this.testRuns = null;
+  }
+
+  async getLatestPassedRun(): Promise<TestRun | null> {
+    const testRuns = await this.getRuns();
+    const validTestRuns = testRuns.filter(
+      (testRun) =>
+        testRun.status === "passed" &&
+        testRun.version === TestRunRepository.VERSION &&
+        !testRun.fromCache,
```
rmarescu, line 192: Test runs from cache don't have AI steps, so they are not useful for executing future runs off them.

packages/shortest/src/core/runner/test-run-repository.ts
```diff
@@ -0,0 +1,411 @@
...
+      }
+
+      this.log.trace("Retention policy applied successfully", {
+        keptRunId: latestPassedRun.runId,
+        totalDeleted: deletedCount,
+      });
+    }
+    // If no passed run, keep only up to MAX_RUNS_PER_TEST most recent runs
+    else {
+      const MAX_RUNS_PER_TEST = 1;
```
rmarescu, line 335: This can be easily changed to keep multiple copies if needed.

packages/shortest/src/core/runner/test-run.ts
```diff
@@ -110,4 +127,46 @@ export class TestRun {
...
+  static fromCacheFile(
+    testCase: TestCase,
+    file: string,
+    cacheEntry: CacheEntry,
+  ): TestRun {
+    const testRun = new TestRun(testCase);
+    testRun.runId = cacheEntry.metadata.runId;
+    testRun.timestamp = cacheEntry.metadata.timestamp;
+    testRun.version =
+      typeof cacheEntry.metadata.version === "string"
```
rmarescu, line 155: Handles the previous version.

packages/shortest/tests/unit/ai/client.test.ts
rmarescu, line 1: AI-generated, haven't reviewed the tests.

rmarescu (commented):
Introduced `TestRun` and `TestRunRepository` to expand the basic functionality of cached data towards baselines.

I'll do a final check tomorrow before merging.

devsargam:
okay, let me know if I need to make any changes

PR #388, opened 2025-03-11
feat: add support for custom Anthropic fetch implementation

r-yoshihara-j:
- Add undici package to handle HTTP/HTTPS proxy configuration
- Implement global proxy dispatcher using EnvHttpProxyAgent
- Update Anthropic provider to use global fetch with proxy support
- Add undici as a direct dependency in package.json

## Motivation
This change enables users to use the Anthropic provider even when behind corporate proxies or other environments requiring proxy configuration. (Note: communications through browsers launched by playwright are not covered by this implementation, so tests with non-localhost addresses might not display properly)

## Testing
All existing tests pass successfully. No additional tests were added as the functionality depends on environment variables.

## Configuration
To use this feature, users need to configure the following environment variables according to their environment:
- HTTP_PROXY
- HTTPS_PROXY
- NO_PROXY
- NODE_EXTRA_CA_CERTS (if needed for custom certificates)

### DIFF
--- a/packages/shortest/src/ai/provider.ts
+++ b/packages/shortest/src/ai/provider.ts
@@ -12,12 +12,7 @@ import { setGlobalDispatcher, EnvHttpProxyAgent } from "undici";
 export const createProvider = (aiConfig: AIConfig): LanguageModelV1 => {
   switch (aiConfig.provider) {
     case "anthropic":
-      if (process.env.HTTPS_PROXY || process.env.HTTP_PROXY) {
-        // Apply proxy-related environment variables to the global fetch function using undici
-        // If using custom CA certificates, set the NODE_EXTRA_CA_CERTS environment variable
-        setGlobalDispatcher(new EnvHttpProxyAgent());
-      }
-      const anthropic = createAnthropic({ apiKey: aiConfig.apiKey, fetch: globalThis.fetch });
+      const anthropic = createAnthropic({ apiKey: aiConfig.apiKey });
       return anthropic(aiConfig.model) as LanguageModelV1;
     default:
       throw new AIError(
--- a/packages/shortest/src/ai/provider.ts
+++ b/packages/shortest/src/ai/provider.ts
@@ -2,7 +2,6 @@ import { createAnthropic } from "@ai-sdk/anthropic";
 import { LanguageModelV1 } from "ai";
 import { AIConfig } from "@/types";
 import { AIError } from "@/utils/errors";
-import { setGlobalDispatcher, EnvHttpProxyAgent } from "undici";
 
 /**
  * Creates a custom AI provider based on the provided configuration.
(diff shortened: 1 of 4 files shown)


packages/shortest/src/ai/provider.ts
```diff
@@ -11,7 +12,12 @@ import { AIError } from "@/utils/errors";
 export const createProvider = (aiConfig: AIConfig): LanguageModelV1 => {
   switch (aiConfig.provider) {
     case "anthropic":
-      const anthropic = createAnthropic({ apiKey: aiConfig.apiKey });
+      if (process.env.HTTPS_PROXY || process.env.HTTP_PROXY) {
```
rmarescu, line 15: I believe network/proxy configuration is better handled at the application level where users have direct access to the infrastructure setup they need (ENVs, etc).

The use case is quite specific and adding it would increase complexity for limited benefit. Instead, Shortest should just pass-through a custom fetch method via the config.
We can propose your solution in the docs for reference, but ultimately the user has the freedom to implement what works for them.

A config that uses the custom `fetch` would look something like this:

```typescript
import type { ShortestConfig } from "@antiwork/shortest";
import { setGlobalDispatcher, EnvHttpProxyAgent } from "undici";

if (process.env.HTTPS_PROXY || process.env.HTTP_PROXY) {
  setGlobalDispatcher(new EnvHttpProxyAgent());
}

export default {
  headless: false,
  baseUrl: "http://localhost:3000",
  testPattern: "**/*.test.ts",
  ai: {
    provider: "anthropic",
    fetch: globalThis.fetch 
  }
} satisfies ShortestConfig;
```

rmarescu (commented):
@r-yoshihara-j thanks for the contribution. I've added a comment, let me know you think! Feel free to implement the suggested change if you're up to it!

packages/shortest/src/ai/provider.ts
r-yoshihara-j: @rmarescu 
Thank you for your comment.
I thought your suggestion was correct, so I reflected the settings in `shortest.config.ts.example` .
Also, I found that the AI SDK uses the global `fetch` function if nothing is passed to the `fetch` argument in `createAnthropic` function, so it works without explicitly specifying it in the arguments. Therefore, I reverted `provider.ts` to its original state.

packages/shortest/src/shortest.config.ts.example
```diff
@@ -1,4 +1,11 @@
 import type { ShortestConfig } from "@antiwork/shortest";
+import { setGlobalDispatcher, EnvHttpProxyAgent } from "undici";
```
rmarescu, line 2: It looks like this solution uses a global proxy setup, which is simpler, and doesn't require any changes for Shortest.

If this works for your needs, then this PR is no longer needed.

To document the usage of the global proxy setup, I recommend [creating a discussion](https://github.com/antiwork/shortest/discussions/new?category=show-and-tell) and share your implementation there, so other can use for inspiration.

r-yoshihara-j: @rmarescu 
As you pointed out, I have [created a discussion](https://github.com/antiwork/shortest/discussions/392).
I will close this pull request.

PR #456, opened 2025-06-24
chore: remove unused sheet component

boogsbunny:
## Background
This PR removes the unused sheet component.

## Changes

- Delete unused `sheet.tsx`

## Testing

Verified with LSP that these files are not used anywhere else.


slavingia:
We don't work on this anymore so these aren't eligible for any bounties; just gumroad Flexile and helper fwiw.

